PR #32, opened 2020-01-14
Implemented robot specific constants

Zinka010:
- Added robot specific constants based on a robot.conf file under `/home/lvuser`
 - Added clear instructions on how to add robot.conf to a robot

Testing performed
Created and verified different values on Mergio

### DIFF
--- a/src/main/java/frc/robot/config/Config.java
+++ b/src/main/java/frc/robot/config/Config.java
@@ -1,4 +1,75 @@
 package frc.robot.config;
 
+import edu.wpi.first.wpilibj.DriverStation;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
 public class Config {
+
+    /**
+     * Instructions for set up of robot.conf file on robot
+     * <p>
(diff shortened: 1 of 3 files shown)


src/main/java/frc/robot/config/Config.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+
+    private static int getRobotId() {
+        if (robotId < 0) {
+            try (BufferedReader reader = Files.newBufferedReader(ROBOT_ID_LOC)) {
+                robotId = Integer.parseInt(reader.readLine());
+            } catch (IOException | NumberFormatException e) {
+                robotId = 0;
+                DriverStation.reportError("Could not find robot configuration file.", false);
```
kevlam2706, line 70: Suggest something like "Could not find robot configuration file. Defaulting to robotId=0" so it's clear in the logs if we're ever debugging. Maybe also output the value of ROBOT_ID_LOC so the logs clearly show where and what it's looking for.

kevlam2706 (approved):
Looks good. Suggested changes per comments.

src/main/java/frc/robot/config/Config.java
Zinka010: In order to do this, we would first need to complete the logging functionality, to begin with. Should that be done before we merge the PR or can I just add it after?

AlexanderTsarapkine (approved):

PR #38, opened 2020-01-18
AnalogSelector class

LaraLim:
SendableBase has been deprecated as of 2020 so this class now implements the Sendable interface. The min and max of each voltage range are now constants for easier adjustability when required. So far the returned index is accurate for 1-6, but for 7 and onward the returned index is unreliable. The voltage ranges for each index should be updated with the selector installed on the 2020 season robot.

### DIFF
--- a/src/main/java/frc/robot/sensors/AnalogSelector.java
+++ b/src/main/java/frc/robot/sensors/AnalogSelector.java
@@ -7,6 +7,9 @@ import edu.wpi.first.wpilibj.smartdashboard.SendableRegistry;
 
 public class AnalogSelector implements Sendable {
 
+    private static final double MODE_ZERO_LOW_VOLTAGE = 0;
+    private static final double MODE_ZERO_HIGH_VOLTAGE = 2.5;
+
     private static final double MODE_ONE_LOW_VOLTAGE = 2.5;
     private static final double MODE_ONE_HIGH_VOLTAGE = 2.75;
 
@@ -44,7 +47,7 @@ public class AnalogSelector implements Sendable {
     private static final double MODE_TWELVE_HIGH_VOLTAGE = 5;
 
     private static final Range[] VOLTAGE_RANGES = {
-        new Range(0, 2.5),
+        new Range(MODE_ZERO_LOW_VOLTAGE, MODE_ZERO_HIGH_VOLTAGE),
         new Range(MODE_ONE_LOW_VOLTAGE, MODE_ONE_HIGH_VOLTAGE), 
         new Range(MODE_TWO_LOW_VOLTAGE, MODE_TWO_HIGH_VOLTAGE),
         new Range(MODE_THREE_LOW_VOLTAGE, MODE_THREE_HIGH_VOLTAGE), 
@@ -56,8 +59,7 @@ public class AnalogSelector implements Sendable {
         new Range(MODE_NINE_LOW_VOLTAGE, MODE_NINE_HIGH_VOLTAGE), 
         new Range(MODE_TEN_LOW_VOLTAGE, MODE_TEN_HIGH_VOLTAGE), 
         new Range(MODE_ELEVEN_LOW_VOLTAGE, MODE_ELEVEN_HIGH_VOLTAGE),
-        new Range(MODE_TWELVE_LOW_VOLTAGE, MODE_TWELVE_HIGH_VOLTAGE), 
-        //new Range(4.6, 5)
+        new Range(MODE_TWELVE_LOW_VOLTAGE, MODE_TWELVE_HIGH_VOLTAGE),
     };
     private final AnalogInput analogInput;
 
--- a/src/main/java/frc/robot/sensors/AnalogSelector.java
+++ b/src/main/java/frc/robot/sensors/AnalogSelector.java
@@ -1,0 +1,120 @@
+package frc.robot.sensors;
+
+import edu.wpi.first.wpilibj.AnalogInput;
+import edu.wpi.first.wpilibj.Sendable;
+import edu.wpi.first.wpilibj.smartdashboard.SendableBuilder;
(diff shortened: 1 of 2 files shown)


Lockscout (approved):
Guess we will have to remeasure for 7 onward when we get the chance, but the code seems fine and it worked when tested.

src/main/java/frc/robot/sensors/AnalogSelector.java
```diff
@@ -0,0 +1,120 @@
...
+    private static final double MODE_TEN_HIGH_VOLTAGE = 4.5;
+
+    private static final double MODE_ELEVEN_LOW_VOLTAGE = 4.5;
+    private static final double MODE_ELEVEN_HIGH_VOLTAGE = 4.6;
+
+    private static final double MODE_TWELVE_LOW_VOLTAGE = 4.6;
+    private static final double MODE_TWELVE_HIGH_VOLTAGE = 5;
+
+    private static final Range[] VOLTAGE_RANGES = {
+        new Range(0, 2.5),
```
georgetzavelas: Do we need this? If so may be worth adding some more constants. Maybe MODE_ZERO?

src/main/java/frc/robot/sensors/AnalogSelector.java
```diff
@@ -0,0 +1,120 @@
...
+        new Range(MODE_FOUR_LOW_VOLTAGE, MODE_FOUR_HIGH_VOLTAGE), 
+        new Range(MODE_FIVE_LOW_VOLTAGE, MODE_FIVE_HIGH_VOLTAGE),
+        new Range(MODE_SIX_LOW_VOLTAGE, MODE_SIX_HIGH_VOLTAGE), 
+        new Range(MODE_SEVEN_LOW_VOLTAGE, MODE_SEVEN_HIGH_VOLTAGE), 
+        new Range(MODE_EIGHT_LOW_VOLTAGE, MODE_EIGHT_HIGH_VOLTAGE),
+        new Range(MODE_NINE_LOW_VOLTAGE, MODE_NINE_HIGH_VOLTAGE), 
+        new Range(MODE_TEN_LOW_VOLTAGE, MODE_TEN_HIGH_VOLTAGE), 
+        new Range(MODE_ELEVEN_LOW_VOLTAGE, MODE_ELEVEN_HIGH_VOLTAGE),
+        new Range(MODE_TWELVE_LOW_VOLTAGE, MODE_TWELVE_HIGH_VOLTAGE), 
+        //new Range(4.6, 5)
```
georgetzavelas: Does this provide future value? If so it would be worth adding a comment, otherwise I would 🔥

src/main/java/frc/robot/sensors/AnalogSelector.java
```diff
@@ -0,0 +1,120 @@
...
+import edu.wpi.first.wpilibj.Sendable;
+import edu.wpi.first.wpilibj.smartdashboard.SendableBuilder;
+import edu.wpi.first.wpilibj.smartdashboard.SendableRegistry;
+
+public class AnalogSelector implements Sendable {
+
+    private static final double MODE_ONE_LOW_VOLTAGE = 2.5;
+    private static final double MODE_ONE_HIGH_VOLTAGE = 2.75;
+
+    private static final double MODE_TWO_LOW_VOLTAGE = 2.75;
```
georgetzavelas: A thought: Do we feel that having the same end point and starting point for two adjacent ranges may lead to confusion if the voltage is exactly at that value? i.e. if the voltage is 2.75, is it Mode 1 or Mode 2?  Unless Ranges are defined `Low <= x < High` as opposed to `Low <= x <= High`.

georgetzavelas:
By the way this cleanup of the previous code implementation is amazing!

src/main/java/frc/robot/sensors/AnalogSelector.java
LaraLim: MODE_ZERO makes sense, I'll add that. I wasn't sure what to do with this range as it is never returned as an index.

src/main/java/frc/robot/sensors/AnalogSelector.java
LaraLim: I was using this as a reference for what to set as the last few (MODE_TWELVE) constants and left it there by accident. Thanks for catching that.

src/main/java/frc/robot/sensors/AnalogSelector.java
LaraLim: The isWithin method used to sort voltage into ranges uses `min <= number && number < max` so I think having the same end point as a starting point for ranges should be ok.

georgetzavelas: 👍

src/main/java/frc/robot/sensors/AnalogSelector.java
kevlam2706: If the physical sensor gets unplugged or damaged, the software might read zero volts on the analog input. (Worth testing on the robot by intentionally unplugging the sensor and seeing what happens). Our code should be able to detect this, so a mode zero is a good idea.

src/main/java/frc/robot/sensors/AnalogSelector.java
kevlam2706: Good thought, and I was thinking the same thing when we saw values flicker between like 6 and 7 on the Shuffleboard when you were testing on Saturday.

kevlam2706 (approved):
Looks good. I like the suggestions made in the comments. It seems like everything should work in theory so I would like us to try and investigate why the selector seems to be erratic after setting 7. I do have a brand new sensor I can bring into the barn if we want to try plugging it in and comparing to see if it gives different results from the other ones.

georgetzavelas (approved):
Nice work!

Zinka010 (approved):
After adding a comment to explain if the max value will read position N or position N+1, it seems good!

PR #50, opened 2020-01-28
Drive Train Turn pid and Pigeon Sensor

### LINKED ISSUE #80: Autonomous: Command turn X degrees
Implement the command to rotate x degrees in autonomous.

Lockscout:
The PID turning command is fully functional for now, unless people come up with things to add in the future. I also added a Pigeon sensor so that it doesn't need re declared every single time we use it.

You supply it the degrees to turn, (negative for left positive for right), and it will turn that many degrees. The pid will need to be re tuned when we get the new robot.

Maps to Kanban issue #11


src/main/java/frc/robot/commands/DrivetrainPIDTurnDelta.java
```diff
@@ -0,0 +1,75 @@
...
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.config.Config;
+import frc.robot.sensors.Pigeon;
+import frc.robot.subsystems.DriveBase;
+
+public class DrivetrainPIDTurnDelta extends CommandBase {
+
+    //Delcare PD variables
+    private double kPgain = 0.01;
+    private double kDgain = 0.0016;
```
nrgill28: These are a great use for FluidConstants!

.idea/modules/2020-2706-Robot-Code.main.iml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
nrgill28: Is this file necessary?

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -56,7 +59,7 @@ private void configureButtonBindings() {
 
     DriveBase.getInstance().setDefaultCommand(driveCommand);
     
-  
+    new JoystickButton(driverStick, XboxValue.XBOX_A_BUTTON.getPort()).whenHeld(new DrivetrainPIDTurnDelta(90));
```
nrgill28: Remove controller bindings used for debug!

src/main/java/frc/robot/commands/DrivetrainPIDTurnDelta.java
```diff
@@ -0,0 +1,75 @@
...
+
+        //Do PD
+        turnThrottle = (targetAngle - currentAngle) * kPgain - (currentAngularRate) * kDgain;
+
+        //Drive
+        drivebase.tankDrive(-turnThrottle, turnThrottle, false);
+    }
+
+    @Override
+    public void end(boolean interrupted) {
```
nrgill28: Remove any methods that you leave unimplemented

nrgill28 (changes requested):

nrgill28 (approved):
Looks good. For anyone wondering, the changes to the FluidConstant class were made to prevent code which only gets the values from FluidConstants from breaking in case the FluidConstant class changes. (Because we might change how FluidConstant works later, but as long as we still implement `Supplier<T>` no code should break!)

src/main/java/frc/robot/subsystems/DriveBase.java
```diff
@@ -32,7 +32,7 @@
     private DriveMode driveMode;
 
     // The drivebase talons
-    private WPI_TalonSRX leftFrontTalon, leftRearTalon, rightFrontTalon, rightRearTalon;
+    public WPI_TalonSRX leftFrontTalon, leftRearTalon, rightFrontTalon, rightRearTalon;
```
georgetzavelas: Curious why you are making these public instead of leaving them private?

.idea/misc.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="ExternalStorageConfigurationManager" enabled="true" />
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_11" default="false" project-jdk-name="11.0.5" project-jdk-type="JavaSDK" />
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_11" project-jdk-name="11" project-jdk-type="JavaSDK" />
```
georgetzavelas: Second time I am seeing this file change by a different person. I wonder if everyone in the team should standardize on their JDK or simply remove this file from being tracked by git? Otherwise this file may be continually changed by everyone's unique environment.

src/main/java/frc/robot/commands/DrivetrainPIDTurnDelta.java
```diff
@@ -0,0 +1,67 @@
...
+import frc.robot.subsystems.DriveBase;
+
+public class DrivetrainPIDTurnDelta extends CommandBase {
+    //Delcare PD variables
+    private Supplier<Double> pGain = Config.DRIVETRAIN_P;
+    private Supplier<Double> dGain = Config.DRIVETRAIN_D;
+
+    //get the drivebase and pigeon
+    private final DriveBase drivebase;
+    private PigeonIMU _pidgey = Pigeon.getPigeon();
```
georgetzavelas: Nitpicks so choose to ignore if you want: To be consistent I would initialize these in the constructor. Also for consistency I would remove the `_` from the variable name unless you are using it to signify something different about this variable vs the others?

src/main/java/frc/robot/config/Config.java
```diff
@@ -63,11 +65,16 @@ private Config() {
...
     public static boolean INVERT_SIDE= robotSpecific(false, false, false);
 
-
     // Timeouts for sending CAN bus commands
     public static final int CAN_TIMEOUT_SHORT = 10;
     public static final int CAN_TIMEOUT_LONG = 100;
 
+    static NetworkTable constantsTable = NetworkTableInstance.getDefault().getTable("constants");
+
+    public static FluidConstant<Double> DRIVETRAIN_P = new FluidConstant<>("DrivetrainP", 0.01d)
```
georgetzavelas: 😻

src/main/java/frc/robot/sensors/Pigeon.java
```diff
@@ -0,0 +1,27 @@
...
+import com.ctre.phoenix.sensors.*;
+import frc.robot.subsystems.DriveBase;
+
+public class Pigeon {
+
+    static PigeonIMU _pidgey;      // Pigeon IMU used to enforce straight drive
+
+    static {
+        //get talon
+        _pidgey = new PigeonIMU(DriveBase.getInstance().leftFrontTalon);
```
georgetzavelas: Have you thought about using a Singleton pattern here instead of initializing this statically? I would recommend a static class for independent calculations, e.g. math routines on passed in variables, but when accessing another instance of a shared resource I think a Singleton has less chance for subtle bugs, e.g. calling this static class from another static variable, since it controls access to the variable through a single instance of an object. Not that this is bad/wrong, just food for thought.

src/main/java/frc/robot/sensors/Pigeon.java
```diff
@@ -0,0 +1,27 @@
...
+        _pidgey.setFusedHeading(0.0, 30);
+    }
+
+    public static PigeonIMU getPigeon() {
+        return _pidgey;
+    }
+
+    public static double getCurrentAngle(){
+        //Gets the current angle
+        PigeonIMU.FusionStatus fusionStatus = new PigeonIMU.FusionStatus();
```
georgetzavelas: Should this be `_pidgey` instead of creating a new object?

georgetzavelas (commented):
Amazing work! Some comments to consider

src/main/java/frc/robot/subsystems/DriveBase.java
kevlam2706: Ah yes. Because he needed to access the talon object in order to pass it to the Pigeon constructor, since that is how you tell the Pigeon that it is attached to a Talon.

My question is - if the Pigeon and the Talon are so strongly interrelated -- AND the Pigeon is physically part of the drivebase -- should the Pigeon be a member of DriveBase?

.idea/misc.xml
kevlam2706: agreed. .gitignore time?

src/main/java/frc/robot/commands/DrivetrainPIDTurnDelta.java
kevlam2706: Or per my previous comment I think there is a case for having the Pigeon instantiated in the DriveBase and then this command would just do a getInstance of it.

src/main/java/frc/robot/sensors/Pigeon.java
kevlam2706: Agreed. Let's get the architecture right from early on so that we start on the right foot and everyone can emulate good practices. The 'double instantiation of a single object' issue is a bug we have seen many many times over the years.

Again I think there's a case for the Pigeon to be managed in the DriveBase. I don't think I like the pattern of having a class (robot.sensors.Pigeon) that is this strongly linked to the DriveBase and in particular the internal specifics of the DriveBase (the fact that it has a left front Talon, this should really be private to the DriveBase. One day we might have a Falcon or NEO based DriveBase :) )

Please consider my suggestion of moving the management of the Pigeon to within the drivebase subsystem.

src/main/java/frc/robot/sensors/Pigeon.java
kevlam2706: Agreed. Let's be very careful about instantiating multiple objects that tie back to a single piece of physical hardware. At worst, there will be a robot code crash (this used to happen to us a lot when we were allocating IOs on the RoboRio). At best this will lead to subtle bugs because one instance will tell the hardware to do X or keep track of Y and another instance will have no such knowledge and your code will get nulls or bad data or worse.

src/main/java/frc/robot/commands/DrivetrainPIDTurnDelta.java
```diff
@@ -0,0 +1,67 @@
...
+        double[] xyz_dps = new double[3];
+        _pidgey.getRawGyro(xyz_dps);
+        //Get z axis angular rate
+        double currentAngularRate = xyz_dps[2];
+
+        //Get current angle
+        currentAngle = Pigeon.getCurrentAngle();
+
+        //Do PD
+        turnThrottle = (targetAngle - currentAngle) * pGain.get() - (currentAngularRate) * dGain.get();
```
kevlam2706, line 78: Huh. Manual PID?

src/main/java/frc/robot/sensors/Pigeon.java
```diff
@@ -0,0 +1,27 @@
...
+public class Pigeon {
+
+    static PigeonIMU _pidgey;      // Pigeon IMU used to enforce straight drive
+
+    static {
+        //get talon
+        _pidgey = new PigeonIMU(DriveBase.getInstance().leftFrontTalon);
+
+        //setup gyro angle finder
+        _pidgey.setFusedHeading(0.0, 30);
```
kevlam2706: The side effect of running this command is that it will reset the pigeon heading.

That's OK if this is the convention we set, but everyone needs to be aware of it, e.g. the other students doing autonomous and vision commands.

The other way to do it is let the Pigeon keep track of its own heading and everyone uses their own variables to offset to what we want.

src/main/java/frc/robot/subsystems/DriveBase.java
kevlam2706: agreed, see my previous comments about rearchitecting where to put the pigeon. If it needs to know about the Talon (which it does) then maybe it should be managed here along with the Talon.

kevlam2706 (approved):
approved but please consider my request for changes to architecture. let's get this right before we let other students loose on it.

src/main/java/frc/robot/sensors/Pigeon.java
nrgill28: No this code is fine. The Pigeon library is very C-style where you instantiate a class and pass it into a function which modifies the class instead of returning a result

kevlam2706: OK, as long as that is clear to everyone. 

This is a general comment but remember that we are writing this code not just for this year but potentially for future years as well, including for rookie students. For that reason I would prioritize things like consistence, readability, clarity, and simplicity over clever tricks and special exceptions.

kevlam2706 (approved):
cool

src/main/java/frc/robot/commands/DrivetrainPIDTurnDelta.java
```diff
@@ -64,6 +75,6 @@ public void execute() {
         turnThrottle = (targetAngle - currentAngle) * pGain.get() - (currentAngularRate) * dGain.get();
 
         //Run motors according to the output of PD
-        drivebase.tankDrive(-turnThrottle, turnThrottle, false);
+        drivebase.tankDrive(-turnThrottle + forwardSpeed, turnThrottle + forwardSpeed, false);
```
nrgill28, line 81: This might be an odd case, but since I think tankDrive clamps it's input to `0 <= x <= 1` you'll run in to some issues when `turnThrottle + forwardSpeed > 1` since the clamp will chop off part of your speed difference.  
I imagine it will only affect how fast you turn though.  

Also I know `A + B` is the same as `B + A` but I think it would look better with the variables swapped around 😛

Lockscout: I think that it would be a rare case, since it would only matter a lot if we are going at max velocity, which I feel like we wouldn't do. I also feel like when used with vision they wouldn't want the robot moving at max speed since that would cause motion blur.

PR #52, opened 2020-01-30
Jacob t

jwt2706:
This is a basic experiment with Shuffleboard. If you want to test it, please use an Xbox 360 controller.

### DIFF
--- a/src/main/java/frc/robot/Robot.java
+++ b/src/main/java/frc/robot/Robot.java
@@ -6,6 +6,7 @@
 /*----------------------------------------------------------------------------*/
 
 package frc.robot;
+
 import edu.wpi.first.wpilibj.Joystick;
 import edu.wpi.first.wpilibj.TimedRobot;
 import edu.wpi.first.wpilibj.smartdashboard.SmartDashboard;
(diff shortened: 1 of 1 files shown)


src/main/java/frc/robot/Robot.java
```diff
@@ -6,139 +6,137 @@
...
-  //private double kI = 0.5;
-  //private double kD = 0.5;
-  //private int setpoint = 0;
-  
-  /**
-   * This function is run when the robot is first started up and should be
-   * used for any initialization code.
-   */
+  private final Timer m_timer = new Timer();
+  private int SecondsLeft = 150;
```
georgetzavelas, line 27: Since this is a constant, how about making it a private static final int? Also the convention for constants in Java is to use capitalization, e.g. GAME_IN_SECONDS?

src/main/java/frc/robot/Robot.java
```diff
@@ -6,139 +6,137 @@
...
+
+      double XvalueL = joystick.getRawAxis(0);
+      double YvalueL = joystick.getRawAxis(1);
+      double XvalueR = joystick.getRawAxis(4);
+      double YvalueR = joystick.getRawAxis(5);
+
+      // Left Joystick
+      SmartDashboard.putNumber("Joystick XL Value", XvalueL);
+      SmartDashboard.putNumber("Joystick YL Value", YvalueL * -1);
+      SmartDashboard.putBoolean("XL Too Slow?", XvalueL < -0.9);
```
georgetzavelas, line 96: What does XL and YL stand for? 
When numbers appear in code that look measured or arbitrary we call them "magic numbers" since their meaning and how they were derived are a mystery to anyone outside of the author. One way to make them more understandable is to assign the number to a descriptive variable name. e.g. public static final double <MEANINGFUL_NAME_OF_VARIABLE> = 0.9

georgetzavelas (commented):
This is a great start Jacob!
- Don't forget to add a more relevant PR subject so this can be more easily referenced in the future.
- Is there a way you can break up the code in teleopPeriodic into some smaller methods instead of inserting all of it into teleopPeriodic?

jwt2706:
Creating Shuffleboard subsystem instead.

src/main/java/frc/robot/Robot.java
jwt2706: Oh I understand. XL meant the x value of the Left joystick, YR is the y value of the Right joystick, and so on... I will make sure to put a better name for the variables.

PR #54, opened 2020-01-30
Added basic intake subsystem and command

nrgill28:
Intake is bound to LEFT BUMPER. Multiple conditions can be set so the intake only operates when they are all met. (e.g. The button is pressed and there is space for another ball)

Also the intake stuff won't do anything if the robot doesn't have an intake.

### DIFF
--- a/src/main/java/frc/robot/RobotContainer.java
+++ b/src/main/java/frc/robot/RobotContainer.java
@@ -11,7 +11,7 @@ import edu.wpi.first.wpilibj.GenericHID;
 import edu.wpi.first.wpilibj.Joystick;
 import edu.wpi.first.wpilibj.XboxController;
 import edu.wpi.first.wpilibj2.command.button.JoystickButton;
-import frc.robot.commands.IntakeCommand;
+import frc.robot.commands.OperatorIntakeCommand;
 import frc.robot.config.Config;
 import frc.robot.subsystems.DriveBase;
 import frc.robot.commands.ArcadeDriveWithJoystick;
@@ -60,7 +60,7 @@ public class RobotContainer {
     DriveBase.getInstance().setDefaultCommand(driveCommand);
     
     // Instantiate the intake command and bind it
-    intakeCommand = new IntakeCommand();
+    intakeCommand = new OperatorIntakeCommand();
     new JoystickButton(driverStick, XboxController.Button.kBumperLeft.value).whenHeld(intakeCommand);
   }
 
--- a/src/main/java/frc/robot/RobotContainer.java
+++ b/src/main/java/frc/robot/RobotContainer.java
@@ -61,7 +61,7 @@ public class RobotContainer {
     
     // Instantiate the intake command and bind it
     intakeCommand = new IntakeCommand();
-    (new JoystickButton(driverStick, XboxController.Button.kBumperLeft.value)).whenHeld(intakeCommand);
+    new JoystickButton(driverStick, XboxController.Button.kBumperLeft.value).whenHeld(intakeCommand);
   }
 
   /**
--- a/src/main/java/frc/robot/RobotContainer.java
+++ b/src/main/java/frc/robot/RobotContainer.java
@@ -10,6 +10,8 @@ package frc.robot;
 import edu.wpi.first.wpilibj.GenericHID;
 import edu.wpi.first.wpilibj.Joystick;
 import edu.wpi.first.wpilibj.XboxController;
+import edu.wpi.first.wpilibj2.command.button.JoystickButton;
+import frc.robot.commands.IntakeCommand;
 import frc.robot.config.Config;
 import frc.robot.subsystems.DriveBase;
 import frc.robot.commands.ArcadeDriveWithJoystick;
@@ -40,6 +42,7 @@ public class RobotContainer {
   private Joystick controlStick;
 
   private Command driveCommand;
+  private Command intakeCommand;
 
 
   /**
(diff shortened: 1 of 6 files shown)


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -52,11 +55,13 @@ private void configureButtonBindings() {
...
 
+    // Instantiate the drive command and bind it
     driveCommand = new ArcadeDriveWithJoystick(driverStick, Config.LEFT_CONTROL_STICK_Y, Config.INVERT_FORWARD, Config.RIGHT_CONTROL_STICK_X, Config.INVERT_SIDE);
-
     DriveBase.getInstance().setDefaultCommand(driveCommand);
     
-  
+    // Instantiate the intake command and bind it
+    intakeCommand = new IntakeCommand();
+    (new JoystickButton(driverStick, XboxController.Button.kBumperLeft.value)).whenHeld(intakeCommand);
```
georgetzavelas: Are the outer brackets needed here?

src/main/java/frc/robot/config/FluidConstant.java
```diff
@@ -88,7 +97,7 @@ public void setValue(T newValue) {
      */
     private void entryUpdated(EntryNotification notification) {
         // Only allow the value to be updated while the robot is disabled
-        if (DriverStation.getInstance().isDisabled()) {
+        if (!DriverStation.getInstance().isDisabled()) {
```
georgetzavelas: How about `DriverStation.getInstance().isEnabled()` to avoid a double negative?

georgetzavelas (commented):
I love the novel design using a HashMap! Some comments

nrgill28:
Good catch! I didn't know there was a `isEnabled` as well as `isDisabled`. Guess I should have read the documentation 😜

georgetzavelas (approved):

src/main/java/frc/robot/config/Config.java
```diff
@@ -68,6 +72,11 @@ private Config() {
     public static final int CAN_TIMEOUT_SHORT = 10;
     public static final int CAN_TIMEOUT_LONG = 100;
 
+    // Fluid constants
+    static NetworkTable constantsTable = NetworkTableInstance.getDefault().getTable("constants");
+    public static FluidConstant<Double> INTAKE_SPEED = new FluidConstant<>("intake-target-speed", 0.25d)
+            .registerToTable(constantsTable);
```
kevlam2706: Is it possible to define this constant inside the Intake subsystem and still have it be treated as a fluid constant?

I guess what I'm asking is if all constants for all subsystems should be defined in one common and global Config class or whether it makes more sense to define them inside the subsystems that use them.

Defining them in one spot makes them easy to find, but that breaks the encapsulation concept of having things contained in Subsystems.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,82 @@
...
+     */
+    private final static IntakeSubsystem INSTANCE = new IntakeSubsystem();
+    
+    // The supplier of the intake speed
+    private Supplier<Double> intakeSpeed;
+    
+    // A dictionary to hold all the conditions for the intake to run
+    private HashMap<String, Boolean> conditions;
+    
+    // The intake motor (if any)
```
kevlam2706: (if any)? I think it's established there will be an intake motor :) Even on robot configs where it's not defined, this is still here.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,82 @@
...
+        intakeSpeed = Config.INTAKE_SPEED;
+        if (Config.INTAKE_MOTOR != -1) {
+            intakeMotor = new VictorSPX(Config.INTAKE_MOTOR);
+        }
+    }
+    
+    /**
+     * Returns the Singleton instance of this IntakeSubsystem. This static method 
+     * should be used -- {@code IntakeSubsystem.getInstance();} -- by external
+     * classes, rather than the constructor to get the instance of this class.
```
kevlam2706: Should there be a constructor defined as private to go along with this, or is that automatic/implicit?

src/main/java/frc/robot/commands/IntakeCommand.java
```diff
@@ -0,0 +1,40 @@
...
+public class IntakeCommand extends CommandBase {
+    
+    private IntakeSubsystem intakeSubsystem;
+    
+    public IntakeCommand() {
+        intakeSubsystem = IntakeSubsystem.getInstance();
+        addRequirements(intakeSubsystem);
+        
+        // Initialize the subsystem with this condition
+        intakeSubsystem.setCondition("button", false);
```
kevlam2706: I wonder if there's a condition label that makes this a bit more clear, such as "activated" or "buttonPressed".

Could this same command theoretically also be used in autonomous or would there need to be a separate AutonomousIntake command?

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+    
+    /**
+     * Sets the condition of the name to the provided value. The intake will only run if all conditions are met
+     * @param name The condition name. e.g. 'Button'
+     * @param condition The condition value.
+     */
+    public void setCondition(String name, boolean condition) {
+        conditions.put(name, condition);
+    }
```
kevlam2706, line 66: This is cool. There is no need for it right now with the code you currently have, but this concept of conditions might be good for catching errors with the subsystems. For example, if there was an encoder in this subsystem, one of the run conditions would be that the encoder was initialized successfully. Same with any other mission critical hardware that cannot be missing or unreadable or Bad Things Happen.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,82 @@
...
+    
+    @Override
+    public void periodic() {
+        // The intakeMotor will be null if the Config entry for it was -1. (Meaning this robot doesn't have an intake)
+        if (intakeMotor == null) return;
+        
+        // Check if all conditions are met.
+        // e.g. The intake button is held down AND there is space for more balls
+        boolean canRun = true;
+        for (var condition : conditions.values()) {
```
kevlam2706: canRun = !conditions.containsValue(false)?

kevlam2706 (approved):
Looking good!

src/main/java/frc/robot/config/Config.java
nrgill28: Yes, it's possible to define them wherever you'd like. I originally put them in the config file because that is how it was done last year, although I would prefer having them in the subsystem classes.

PR #55, opened 2020-01-30
Remove IntelliJ's .idea folder from the project

nrgill28:
As 'useful' as it is, I think we can do without it as it changes the file contents and auto-generates useless files all the time.


georgetzavelas (approved):

kevlam2706 (approved):
Cool. Good idea.

PR #56, opened 2020-02-01
Finished First Iteration of Subsystem Conditions

nrgill28:
So this is an interesting one. I moved the conditions part of the IntakeSubsystem into it's own set of classes so that all subsystems can benefit from them without having to re-write that code.

To summarize the new system: 
- A subsystem creates it's own instance of SubsystemConditionManager
- Commands and other subsystems can create conditions for that manager
- The conditions are created through `SubsystemConditionManager.createCondition` and are automatically registered so no extra code is required.
- The command or other subsystem can now enable and disable it's condition like normal

For an example of how these work, I've modified the intake subsystem and the operator intake command to use this system.

This is very much a first iteration of the code, it's not perfect but I'd like to put it out there and see what you guys think before I go any further. Aside from the regular code review, I'm also looking for feedback on a couple extra things:  
- The naming of basically everything. I'm bad at names, won't pretend I'm not. If you have a better name for a variable or class LMK.
- The placement of the two files. I put the SubsystemConditionManager and SubsystemCondition classes in the subsystems folder mainly because there wasn't already a better place to put it; but since they're not subsystems themselves I think it would be better to move them somewhere else.


src/main/java/frc/robot/commands/OperatorIntakeCommand.java
```diff
@@ -2,25 +2,27 @@
...
     public OperatorIntakeCommand() {
         intakeSubsystem = IntakeSubsystem.getInstance();
         addRequirements(intakeSubsystem);
         
-        // Initialize the subsystem with this condition
-        intakeSubsystem.setCondition("buttonPressed", false);
-        intakeSubsystem.setConditionActive("buttonPressed", IntakeSubsystem.Condition.ONLY_TELEOP);
+        // Initialize the condition
+        condition = intakeSubsystem.getConditionManager()
+                .createCondition("buttonPressed", SubsystemCondition.ConditionOptions.TELEOP, false);
```
georgetzavelas: Can we make "buttonPressed" into a constant?

src/main/java/frc/robot/subsystems/SubsystemCondition.java
```diff
@@ -0,0 +1,79 @@
+package frc.robot.subsystems;
+
+import java.util.EnumSet;
+
+public class SubsystemCondition {
```
georgetzavelas: How about `SubsystemOperation`?

src/main/java/frc/robot/subsystems/SubsystemCondition.java
```diff
@@ -0,0 +1,79 @@
+package frc.robot.subsystems;
+
+import java.util.EnumSet;
+
+public class SubsystemCondition {
+    
+    /**
+     * A list of the possible options conditions can have
+     */
+    public enum ConditionOptions {
```
georgetzavelas: How about `State`?
Then when invoking this elsewhere it will look like `SubsystemOperation.State.TELEOP`

src/main/java/frc/robot/subsystems/SubsystemCondition.java
```diff
@@ -0,0 +1,79 @@
...
+     * A list of the possible options conditions can have
+     */
+    public enum ConditionOptions {
+        ENABLED_FOR_TELEOP,
+        ENABLED_FOR_AUTO;
+        
+        // Enum sets of the combinations for the options
+        public static EnumSet<ConditionOptions> TELEOP = EnumSet.of(ENABLED_FOR_TELEOP);
+        public static EnumSet<ConditionOptions> AUTO = EnumSet.of(ENABLED_FOR_AUTO);
+        public static EnumSet<ConditionOptions> BOTH = EnumSet.of(ENABLED_FOR_TELEOP, ENABLED_FOR_AUTO);
```
georgetzavelas: BOTH is requires me to have more context into what this means. How about something more explicit like `TELEP_AND_AUTO`?

src/main/java/frc/robot/subsystems/SubsystemConditionManager.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.ArrayList;
+import java.util.EnumSet;
+
+/**
+ * A SubsystemConditionManager manages a list of conditions required for this subsystem to run.
+ * Commands and other subsystems can register conditions and the owner of the manager can check if they are satisfied.
+ */
+public class SubsystemConditionManager {
+    // A list of all the conditions for this condition manager
+    private ArrayList<SubsystemCondition> conditions = new ArrayList<>();
```
georgetzavelas: Thoughts on using a HashSet here so you avoid duplicates?

georgetzavelas (approved):
youre-awesome

Zinka010 (approved):

Zinka010:
thumbs-up

Issue #33, opened 2020-01-18
Arm Subsystem and Operator Control


nicojopr:
Software Requirements Arm:
 
Arm must keep its height(software must prevent the elevator from descending when raised) in up to 5 different positions:

- down for intake
- fully up to start/climb, 
- slightly higher than ground for defense
- inside frame perimeter for more protection
- control panel (stretch goal))

Definition of Done:
Production software for the arm subsystem in the robot code, including:
- interface definitions, motors, sensors
- commands
- etc..
- tested independently
- merged and integrated with 2020 robot code in github
- ready for integration

kevlam2706:
I believe this issue is done or close to done. I have added details for the next increment in issue #78 .

PR #57, opened 2020-02-01
Created Arm Subsystem

Zinka010:
- Set up all necessary talons and parameters in order to use the robot arm
 - All PID and Setpoint values will need to be tuned to the real robot
 - Will need to add MoveArmToSetpoint command


src/main/java/frc/robot/subsystems/Arm.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * Initialize the current instance if one does not already exist
+     */
+    public static void init() {
+        if (currentInstance == null) {
+            currentInstance = new Arm();
+        }
+    }
```
nrgill28: You could replace all of this init stuff with a simple  
`private static ArmSubsystem INSTANCE = new ArmSubsystem();`  
at the top of your file. They work almost identically, although this method is already used in the intake subsystem and I think we should standardize how we do it.

src/main/java/frc/robot/subsystems/Arm.java
```diff
@@ -0,0 +1,118 @@
...
+    // TODO Change placeholder values to actual limits
+    private static final int forwardLimit = 100000;
+    private static final int reverseLimit = 0;
+
+    private static Arm currentInstance;
+    // TODO Change placeholder value to robotSpecific
+    WPI_TalonSRX armTalon;
+    ErrorCode errorCode;
+
+    private Arm() {
```
nrgill28: I'd suffix this with `Subsystem` just for clarity

src/main/java/frc/robot/subsystems/DriveBase.java
```diff
@@ -58,6 +58,9 @@ private DriveBase() {
 
     }
 
+    /**
```
georgetzavelas, line 61: Thank you for documenting all of these!!

src/main/java/frc/robot/subsystems/Arm.java
```diff
@@ -0,0 +1,118 @@
...
+
+
+    @Override
+    public void periodic() {
+        super.periodic();
+
+        SmartDashboard.putNumber("Arm Motor Ticks", armTalon.getSelectedSensorPosition(0));
+
+        if (errorCode.value == 0) {
+            SmartDashboard.putString("Arm Talon Error", "Working Fine");
```
nrgill28: Does it make sense to call it an 'error' if it's working fine? Maybe update this to "Arm Status"

src/main/java/frc/robot/config/FluidConstant.java
```diff
@@ -76,14 +76,6 @@ public T getValue() {
         return this.value;
     }
     
-    /**
-     * Get method for implementing the Supplier interface
-     * @return The current value of this constant
-     */
-    public T get() {
```
georgetzavelas, line 83: Any reason that this was removed?

src/main/java/frc/robot/subsystems/DriveBase.java
```diff
@@ -58,6 +58,9 @@ private DriveBase() {
...
+     * Initialize the current DriveBase instance
+     */
     public static void init() {
```
nrgill28, line 64: Again with the standardize on the initialization method. See the IntakeSubsystem for an example

src/main/java/frc/robot/Robot.java
```diff
@@ -21,100 +21,105 @@
...
+    /**
+     * This function is run when the robot is first started up and should be used for any
+     * initialization code.
+     */
+    @Override
+    public void robotInit() {
+        // Instantiate our RobotContainer.  This will perform all our button bindings, and put our
+        // autonomous chooser on the dashboard.
+        DriveBase.init();
+        Arm.init();
```
nrgill28: This way of initializing the subsystems is alright, although it would make more sense to just do it in the subsystem file itself. See the IntakeSubsystem for an example

src/main/java/frc/robot/config/Config.java
```diff
@@ -68,13 +70,23 @@ private Config() {
     public static boolean INVERT_FORWARD = robotSpecific(true, true, true);
     public static boolean INVERT_SIDE= robotSpecific(false, false, false);
 
+    public static boolean INVERT_ARM_TALON = robotSpecific(false, false, false);
+
+    public static int ARM_ALLOWABLE_CLOSED = 4096;
```
georgetzavelas: ARM_ALLOWABLE_CLOSED....? ticks, timeout in seconds, degrees of rotation?

src/main/java/frc/robot/config/FluidConstant.java
```diff
@@ -76,14 +76,6 @@ public T getValue() {
         return this.value;
     }
     
-    /**
-     * Get method for implementing the Supplier interface
-     * @return The current value of this constant
-     */
-    public T get() {
-        return this.value;
-    }
```
nrgill28, line 85: This is required for the FluidConstant to implement the Supplier interface

src/main/java/frc/robot/config/Config.java
```diff
@@ -68,13 +70,23 @@ private Config() {
...
+
     // Timeouts for sending CAN bus commands
     public static final int CAN_TIMEOUT_SHORT = 10;
     public static final int CAN_TIMEOUT_LONG = 100;
 
+    // PIDF values for the arm
+    public static double ARM_P = robotSpecific(0.2);
```
georgetzavelas: Nitpick: I understand what these mean as I see them together. For clarity you may want to consider ARM_PID_P_VALUE, etc.

src/main/java/frc/robot/subsystems/Arm.java
```diff
@@ -0,0 +1,118 @@
...
+import com.ctre.phoenix.motorcontrol.StatusFrame;
+import com.ctre.phoenix.motorcontrol.can.WPI_TalonSRX;
+import edu.wpi.first.wpilibj.smartdashboard.SmartDashboard;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.config.Config;
+
+public class Arm extends SubsystemBase {
+
+    // TODO Change placeholder values to actual limits
+    private static final int forwardLimit = 100000;
```
georgetzavelas: May want to consider naming these in caps to denote that they are constants when used. Also adding TICKS to the name would clarify what limit they represent.

nrgill28 (changes requested):
A few nitpicky things and one issue. Most of this is just standardizing the code across all subsystems.

src/main/java/frc/robot/subsystems/Arm.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+
+    @Override
+    public void periodic() {
+        super.periodic();
+
+        SmartDashboard.putNumber("Arm Motor Ticks", armTalon.getSelectedSensorPosition(0));
+
+        if (errorCode.value == 0) {
```
georgetzavelas: Is anything updating this value?

georgetzavelas (commented):
Nice work!  Some comments

Best viewed with `Hide whitespace changes`
Screen Shot 2020-02-01 at 1 07 53 PM

src/main/java/frc/robot/subsystems/DriveBase.java
Zinka010: I'll deal with this in a separate PR as this is unrelated to the current one.

src/main/java/frc/robot/config/FluidConstant.java
Zinka010: I removed this since it was a duplicate which was somehow created during a merge.

src/main/java/frc/robot/config/FluidConstant.java
Zinka010: I removed this since it was a duplicate that was somehow created during a merge.

src/main/java/frc/robot/subsystems/Arm.java
Zinka010: This value is not being periodically updated no. Should it be moved to the constructor then?

nrgill28 (approved):
Looks good.

src/main/java/frc/robot/subsystems/Arm.java
georgetzavelas: No just wondered. All good.

georgetzavelas (approved):

src/main/java/frc/robot/subsystems/ArmSubsystem.java
```diff
@@ -0,0 +1,109 @@
...
+
+        // Config factory default to clear out any lingering values
+        armTalon.configFactoryDefault();
+
+        // Allow the arm to be moved easily when disabled
+        armTalon.setNeutralMode(NeutralMode.Coast);
+
+        // Setup the talon, recording the error code
+        errorCode = armTalon.configSelectedFeedbackSensor(FeedbackDevice.CTRE_MagEncoder_Relative,
+                0, Config.CAN_TIMEOUT_SHORT);
```
kevlam2706, line 38: I see what you do with errorCode later on. Good start.

I suggest that we incorporate the "conditions" framework that Nathan set up in the IntakeSubsystem and use that as a template. One of the conditions for every subsystem should be that all of its internal health checks pass. So a failing errorCode here could set a "encoderHealthy" condition to FALSE. Then when commands try to operate the arm it would not try to run the motor and we could dump out error logs indicating why.

kevlam2706 (approved):

src/main/java/frc/robot/subsystems/ArmSubsystem.java
Zinka010: That sounds like a good idea, I'll implement it tonight

nicojopr:
Associated with task #33

PR #60, opened 2020-02-04
Moved most of the conditional logic into an abstract class

nrgill28:
Subsystems can now extend this class instead of managing their own manager. All the functionality is the same as before, except now the subsystem defines the conditions and commands and other subsytems can only get a reference to it.


Zinka010 (approved):
Seal of approval

georgetzavelas (approved):

PR #61, opened 2020-02-04
Shuffleboard Subsystem and Enable

### LINKED ISSUE #22: Prototyping: Feedback to Driver
Investigate ways to provide feedback to the driver via robot software, whether to controller or to bling via network tables, etc..

jwt2706:
There is a Shuffleboard subsystem and it can be enabled.


src/main/java/frc/robot/Robot.java
```diff
@@ -8,10 +8,9 @@
 package frc.robot;
 
 import edu.wpi.first.wpilibj.TimedRobot;
-import edu.wpi.first.wpilibj.smartdashboard.SmartDashboard;
 import edu.wpi.first.wpilibj2.command.Command;
 import edu.wpi.first.wpilibj2.command.CommandScheduler;
-import frc.robot.config.Config;
+//import frc.robot.subsystems.ArmSubsystem;  // could not be resolved... will be fixed if needed
```
georgetzavelas: If this can't be resolved we should remove it.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -104,3 +104,4 @@ public void periodic() {
     }
 }
 
+//hi
```
georgetzavelas, line 58: hello!

src/main/java/frc/robot/subsystems/Shuffleboard.java
```diff
@@ -0,0 +1,109 @@
...
+import edu.wpi.first.wpilibj.Joystick;
+import edu.wpi.first.wpilibj.smartdashboard.SmartDashboard;
+import edu.wpi.first.wpilibj.GenericHID.RumbleType;
+import edu.wpi.first.cameraserver.CameraServer;
+import edu.wpi.first.wpilibj.Timer;
+
+public class Shuffleboard extends SubsystemBase {
+  private final Joystick joystick = new Joystick(0);
+  private final Timer m_timer = new Timer();
+  private int SecondsLeft = 150;
```
georgetzavelas: How about this
```
public static final int MAX_SECONDS_IN_MATCH = 150;
private int secondsLeft = MAX_SECONDS_IN_MATCH;
```

When numbers are used in code that seem to be ambiguous we call them "magic numbers". By assigning the number to a clearly labelled variable name it makes the code more understandable without even requiring comments :)

src/main/java/frc/robot/subsystems/Shuffleboard.java
```diff
@@ -0,0 +1,109 @@
...
+
+  private static Shuffleboard INSTANCE = new Shuffleboard();
+
+  /**
+   * Creates a new Shuffleboard.
+   */
+  private Shuffleboard() {
+    CameraServer.getInstance().startAutomaticCapture();
```
georgetzavelas, line 30: This was useful for testing but are we sure we want to start automatic capture here?

src/main/java/frc/robot/subsystems/Shuffleboard.java
```diff
@@ -0,0 +1,109 @@
...
+import edu.wpi.first.wpilibj.GenericHID.RumbleType;
+import edu.wpi.first.cameraserver.CameraServer;
+import edu.wpi.first.wpilibj.Timer;
+
+public class Shuffleboard extends SubsystemBase {
+  private final Joystick joystick = new Joystick(0);
+  private final Timer m_timer = new Timer();
+  private int SecondsLeft = 150;
+
+  private static Shuffleboard INSTANCE = new Shuffleboard();
```
georgetzavelas: For a Singleton pattern you would want to have this instantiation inside of the private constructor. Here is an example of a Singleton pattern: https://github.com/georgetzavelas/guava-examples/blob/master/src/com/design/pattern/creational/singleton/Singleton.java

src/main/java/frc/robot/subsystems/Shuffleboard.java
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Creates a new Shuffleboard.
+   */
+  private Shuffleboard() {
+    CameraServer.getInstance().startAutomaticCapture();
+    m_timer.start();
+    SmartDashboard.putNumber("Time Left", SecondsLeft);
```
georgetzavelas: I would remove this from the constructor and put it in the `periodic` function since we shouldn't be manipulating our object in the constructor as a good practice.

src/main/java/frc/robot/subsystems/Shuffleboard.java
```diff
@@ -0,0 +1,109 @@
...
+  private int SecondsLeft = 150;
+
+  private static Shuffleboard INSTANCE = new Shuffleboard();
+
+  /**
+   * Creates a new Shuffleboard.
+   */
+  private Shuffleboard() {
+    CameraServer.getInstance().startAutomaticCapture();
+    m_timer.start();
```
georgetzavelas, line 31: I see you are already starting the timer in the periodic function. That would be the best place to start it assuming that periodic is always being called from the Robot code.

src/main/java/frc/robot/subsystems/Shuffleboard.java
```diff
@@ -0,0 +1,109 @@
...
+      boolean buttonValueB; // B is 2
+      boolean buttonValueX; // X is 3
+      boolean buttonValueY; // Y is 4
+
+      double XvalueL = joystick.getRawAxis(0);
+      double YvalueL = joystick.getRawAxis(1);
+      double XvalueR = joystick.getRawAxis(4);
+      double YvalueR = joystick.getRawAxis(5);
+
+      // Left Joystick
```
georgetzavelas, line 65: What about putting all of these in a function called `getLeftJoystickInfo`?

src/main/java/frc/robot/subsystems/Shuffleboard.java
```diff
@@ -0,0 +1,109 @@
...
+      SmartDashboard.putNumber("Joystick XL Value", XvalueL);
+      SmartDashboard.putNumber("Joystick YL Value", YvalueL * -1);
+      SmartDashboard.putBoolean("XL Too Slow?", XvalueL < -0.9);
+      SmartDashboard.putBoolean("XL Too Fast?", XvalueL > 0.9);
+      SmartDashboard.putBoolean("YL Too Fast?", YvalueL < -0.9);
+      SmartDashboard.putBoolean("YL Too Slow?", YvalueL > 0.9);
+
+      // Right Joystick
```
georgetzavelas, line 73: What about putting all of these in a function called `getRightJoystickInfo`?

georgetzavelas (commented):
Amazing work!!!!
I have some comments for your consideration.

src/main/java/frc/robot/subsystems/Shuffleboard.java
kevlam2706: If we want the video camera streamed to the driver station Shuffleboard widget, this call needs to happen, somewhere, once. It doesn't have to happen here though.

Also - this is something we need to coordinate with the Vision team. Are we taking care of the driver cameras by streaming video from webcams plugged into the RoboRio, or are the cameras going to be plugged into the Pis and taken care of external to the RoboRio? Jacob has demonstrated the ability to get pretty nice frame rates in his testing. We should demo this functionality to the vision team and see what they think.

src/main/java/frc/robot/subsystems/Shuffleboard.java
```diff
@@ -0,0 +1,109 @@
...
+
+  public static Shuffleboard getINSTANCE() {
+    return INSTANCE;
+  }
+
+  @Override
+  public void periodic() {
+
+    // This method will be called once per scheduler run
+    // shift alt F
```
kevlam2706, line 42: Suggest you remove this comment as it isn't relevant to your code. Maybe write it down in your notebook or somewhere if you want to remember. I'm glad you're using autoformat, I think others should use it too.

src/main/java/frc/robot/subsystems/Shuffleboard.java
```diff
@@ -0,0 +1,109 @@
...
+      SmartDashboard.putBoolean("A", buttonValueA);
+      buttonValueB = joystick.getRawButton(2); // Button B
+      SmartDashboard.putBoolean("B", buttonValueB);
+      buttonValueX = joystick.getRawButton(3); // Button X
+      SmartDashboard.putBoolean("X", buttonValueX);
+      buttonValueY = joystick.getRawButton(4); // Button Y
+      SmartDashboard.putBoolean("Y", buttonValueY);
+
+    } else {
+      // game over signal can be added here
```
kevlam2706, line 106: need to coordinate this with Wei and David since I think they are tasked with setting up the game over signal.

src/main/java/frc/robot/subsystems/Shuffleboard.java
```diff
@@ -0,0 +1,109 @@
...
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+
+import edu.wpi.first.wpilibj.Joystick;
+import edu.wpi.first.wpilibj.smartdashboard.SmartDashboard;
+import edu.wpi.first.wpilibj.GenericHID.RumbleType;
+import edu.wpi.first.cameraserver.CameraServer;
+import edu.wpi.first.wpilibj.Timer;
+
+public class Shuffleboard extends SubsystemBase {
+  private final Joystick joystick = new Joystick(0);
```
kevlam2706, line 19: Does grabbing a joystick instance here interfere with the existing joystick instances that are elsewhere in the code? Did you test this all on a working, driving, robot platform?

kevlam2706 (approved):
Cool stuff Jacob!

This code is all great "proof of concept" that we can track inputs, display them on screen, and do things with them like turn on joystick rumble. Pretty soon we will know from the drive team what they actually want to see, so we can replace this code with "the real thing". :smile:

src/main/java/frc/robot/subsystems/Shuffleboard.java
jwt2706: I don't think it does, I tested it on Monday (on Mergio) and it worked fine. However, maybe it's getting the joystick value twice (for shuffleboard and the robot movement)... i'm guessing we can remove one of them to optimize the performance ...¯\_(ツ)_/¯

src/main/java/frc/robot/subsystems/Shuffleboard.java
jwt2706: Will do!

src/main/java/frc/robot/subsystems/Shuffleboard.java
jwt2706: Understood

src/main/java/frc/robot/subsystems/Shuffleboard.java
jwt2706: If that's easier to understand while reading the code, then so it be

src/main/java/frc/robot/subsystems/Shuffleboard.java
jwt2706: If that's easier to understand while reading the code, then so it be

jwt2706: Since I'm waiting on the drive team, I don't think it's worth changing that now. It might just get deleted anyway, but if we do end up using it, I will do those changes.

src/main/java/frc/robot/config/FluidConstant.java
```diff
@@ -75,14 +75,8 @@ public T get() {
     public T getValue() {
         return this.value;
     }
+
     
-    /**
-     * Get method for implementing the Supplier interface
```
nicojopr: The duplicate get method has been resolved in Justin's commit and you have no other changes in FluidConstants. Since this is showing as a merge conflict on your pull request I would just remove this file from your pull request.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
nicojopr, line 58: :) IntakeSubsystem could also be removed from your pull request

nicojopr (changes requested):
Apply comments as discussed and remove unnecessary files from the pull request

nicojopr (approved):
Thanks for updating

jwt2706:
Finally...done

PR #62, opened 2020-02-05
Game over signal issue #21

### LINKED ISSUE #21: Prototyping: Game Over Signal
aka End Game Signal
controls team would like to know if there is a way we can signal the Pi's to
start shutdown procedure at the end of a match
does FMS send robot code an indicator that the match is over? automatically send something at 5 seconds left? Driver control?
can we send a signal out (i.e. on networktables) once this signal is received?
prefer "end match" over "robot disable" or else we'll have the Pis coming up and down all the time as we practice

davidliu-2706:
This code change sends a shut down signal to the control system and the vision system when the game is over. It also sends a start-up signal to the vision system when the game begins. 
Network tables are used to communicate.

### DIFF
--- a/src/main/java/frc/robot/Robot.java
+++ b/src/main/java/frc/robot/Robot.java
@@ -28,11 +28,13 @@ public class Robot extends TimedRobot {
   private RobotContainer m_robotContainer;
 
   //network table for vision control
-  private VisionCtrlNetTable VisionControlNetTable;
+  private VisionCtrlNetTable m_VisionControlNetTable;
   //network table for control systems control
-  private ControlCtrlNetTable ControlCtrlNetTable;
+  private ControlCtrlNetTable m_ControlCtrlNetTable;
   //flag to indicate from the teleop mode
-  private Boolean bFromTeleMode;
+  private Boolean m_bFromTeleMode;
+  //flag to indicate the real match
+  private Boolean m_bRealMatch;
 
   /**
    * This function is run when the robot is first started up and should be used for any
@@ -47,12 +49,12 @@ public class Robot extends TimedRobot {
     m_robotContainer = new RobotContainer();
     
     //create a vision control table
-    VisionControlNetTable  = new VisionCtrlNetTable();
+    m_VisionControlNetTable  = new VisionCtrlNetTable();
     //create a control system control table
-    ControlCtrlNetTable = new ControlCtrlNetTable();
+    m_ControlCtrlNetTable = new ControlCtrlNetTable();
     //set to false
-    bFromTeleMode = false;
-
+    m_bFromTeleMode = false;
+    m_bRealMatch = false;
   }
 
   /**
@@ -77,12 +79,12 @@ public class Robot extends TimedRobot {
    */
   @Override
   public void disabledInit() {
-    if (isRealMatch() && bFromTeleMode == true) 
+    if (m_bRealMatch == true && m_bFromTeleMode == true) 
     {
       // if in a real match and from teleop mode
       // Write to the network table the shut down signal.
-      VisionControlNetTable.shutDownVision();
-      ControlCtrlNetTable.shutDownControl();
+      m_VisionControlNetTable.shutDownVision();
+      m_ControlCtrlNetTable.shutDownControl();
 
       System.out.println("Driver Station Disabled");      
     }
@@ -106,9 +108,10 @@ public class Robot extends TimedRobot {
     }
 
     //send start up signal to vision team
-    VisionControlNetTable.startUpVision();
+    m_VisionControlNetTable.startUpVision();
 
-    bFromTeleMode = false;
+    m_bFromTeleMode = false;
+    m_bRealMatch = isRealMatch();
   }
 
   /**
@@ -129,7 +132,7 @@ public class Robot extends TimedRobot {
     }
     
     //set teleop mode to true
-    bFromTeleMode = true;
+    m_bFromTeleMode = true;
 
   }
 
@@ -146,7 +149,7 @@ public class Robot extends TimedRobot {
     CommandScheduler.getInstance().cancelAll();
 
     //set teleop mode to false
-    bFromTeleMode = false;
+    m_bFromTeleMode = false;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/main/java/frc/robot/Robot.java
```diff
@@ -25,6 +27,13 @@
 
   private RobotContainer m_robotContainer;
 
+  //network table for vision control
+  private VisionCtrlNetTable m_VisionCtrlNetTable;
+  //network table for control systems control
+  private ControlCtrlNetTable m_ControlCtrlNetTable;
+  //flag to indicate from the teleop mode
+  private Boolean m_bFromTeleMode;
```
georgetzavelas: This variable isn't needed since you can simply call `DriverStation.getInstance().isOperatorControl()` to know if you are in TeleOp.

src/main/java/frc/robot/Robot.java
```diff
@@ -36,6 +45,13 @@ public void robotInit() {
     DriveBase.init();
 
     m_robotContainer = new RobotContainer();
+    
+    //create a vision control table
```
georgetzavelas, line 51: It is nice that you are documenting your code but for simple straightforward concepts they don't give any extra value to the reader so it may be worth removing theses ones.

src/main/java/frc/robot/nettables/ControlCtrlNetTable.java
```diff
@@ -0,0 +1,37 @@
...
+import edu.wpi.first.networktables.NetworkTable;
+import edu.wpi.first.networktables.NetworkTableEntry;
+import edu.wpi.first.networktables.NetworkTableInstance;
+
+public class ControlCtrlNetTable {
+
+    //For NetworkTables
+    //Get the default instance of NetworkTables that was created automatically
+    //when your program starts
+    private NetworkTableInstance inst = NetworkTableInstance.getDefault();
```
georgetzavelas: nitpick: maybe `instance` instead of `inst`?

georgetzavelas (commented):
Great work David!!!
I left some comments for your consideration.

src/main/java/frc/robot/Robot.java
```diff
@@ -25,6 +27,13 @@
 
   private RobotContainer m_robotContainer;
 
+  //network table for vision control
+  private VisionCtrlNetTable m_VisionCtrlNetTable;
```
Zinka010: Kind of a nitpick but to maintain consistency with the rest of the codebase these should be defined without underscores. For example `VisionControlNetTable visionControlNetTable`

src/main/java/frc/robot/Robot.java
```diff
@@ -61,6 +77,16 @@ public void robotPeriodic() {
...
   @Override
   public void disabledInit() {
+    if (isRealMatch() && m_bFromTeleMode == true) 
+    {
+      // if in a real match and from teleop mode
+      // Write to the network table the shut down signal.
+      m_VisionCtrlNetTable.shutDownVision();
+      m_ControlCtrlNetTable.shutDownControl();
+
+      System.out.print("Driver Station Disabled\n");
```
Zinka010: Why not just use `System.out.println` and remove the `\n`?

src/main/java/frc/robot/nettables/ControlCtrlNetTable.java
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Writes a shut down signal to the controls network table.
+     *
+     */
+    public void shutDownControl()
+    {
+        ControlShutDownEntry.setBoolean(true);
+    }
+
```
Zinka010: nitpick: remove useless whitespace

Zinka010 (changes requested):
Overall it's really good, but there's some minor stuff I'd like you to change.

src/main/java/frc/robot/Robot.java
```diff
@@ -25,6 +27,13 @@
 
   private RobotContainer m_robotContainer;
 
+  //network table for vision control
+  private VisionCtrlNetTable VisionControlNetTable;
```
Zinka010: Sorry to have lead you down the wrong path, but for the variable name by convention we use a lower case letter.

Zinka010 (changes requested):
One small thing but overall it's so much better!

src/main/java/frc/robot/Robot.java
davidliu-2706: I chose not to change this to the suggestion because, at the end of the match, the connection between the robot and drive station is possibly terminated, resulting in disconnection and the function not working. Also, when the robot is disabled, it is in disabled mode, so it won't call that code anymore.
We need the teleop mode before the robot is disabled because there might be a short disabled time in between the autonomous mode and the teleop mode.

src/main/java/frc/robot/Robot.java
```diff
@@ -25,6 +27,15 @@
 
   private RobotContainer m_robotContainer;
 
+  //network table for vision control
+  private VisionCtrlNetTable m_VisionControlNetTable;
```
Zinka010: Why is this back to the m_variable naming scheme?

src/main/java/frc/robot/Robot.java
georgetzavelas: I am confused about this. Does it matter if this happens at the end of the match? The robot is disabled at the end of the match and the match is over. The DriverStation is used to operate the Robot by the DriveTeam so it should be trusted IMHO. 

Can you tell me where you have found information that leaves you with this opinion as it was the first time I have heard of this situation, not that I don't believe you, just hadn't heard from anyone over the years that this was a concern.

kevlam2706 (approved):

src/main/java/frc/robot/Robot.java
```diff
@@ -20,104 +23,148 @@
...
-        // and running subsystem periodic() methods.  This must be called from the robot's periodic
-        // block in order for anything in the Command-based framework to work.
-
-        CommandScheduler.getInstance().run();
+  private Command m_autonomousCommand;
+
+  private RobotContainer m_robotContainer;
+
+  //network table for vision control
+  private VisionCtrlNetTable visionControlNetTable;
```
Zinka010, line 31: 🥇

Zinka010 (approved):

georgetzavelas:
@davidliu-2706 don't know if you saw my follow up comment. just wondering if we can touch base on it IRL?

nicojopr:
Maps to task #21

PR #65, opened 2020-02-06
Added Logging

Lockscout:
Ive done the research about how we should log stuff.

The example is in the wiki.

### DIFF
--- a/src/main/java/frc/robot/commands/DrivetrainPIDTurnDelta.java
+++ b/src/main/java/frc/robot/commands/DrivetrainPIDTurnDelta.java
@@ -8,9 +8,11 @@
 package frc.robot.commands;
 
 import java.util.function.Supplier;
+import java.util.logging.Logger;
 
 import com.ctre.phoenix.sensors.PigeonIMU;
 import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.config.Config;
 import frc.robot.subsystems.DriveBase;
 
 public class DrivetrainPIDTurnDelta extends CommandBase {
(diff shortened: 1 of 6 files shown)


src/main/java/frc/robot/commands/DrivetrainPIDTurnDelta.java
```diff
@@ -34,6 +36,9 @@
     //Initiate forwardSpeed
     private double forwardSpeed;
 
+    private Logger logger = Logger.getLogger("DTPIDDelta");
```
kevlam2706, line 39: Should the string we pass to Logger be the same as our class name? e.g. "DrivetrainPIDTurnDelta"?

src/main/java/frc/robot/config/Config.java
```diff
@@ -32,6 +36,19 @@
 
     private static final Path ROBOT_ID_LOC = Paths.get(System.getProperty("user.home"), "robot.conf");
 
+    public static FileHandler logFileHandler;
+
+    static {
+        try {
+            String logFilename = new SimpleDateFormat("'Robotlog_'yyyy'-'MM'-'dd'_'HH'-'mm'-'ss'.txt'").format(new Date());
```
kevlam2706, line 43: Cool. We'll need to remember that the logging code is writing files to the roborio file system, and check in every now and then to make sure we're not using up all of its disk space (there isn't much to begin with).

In future we could add logic to auto-delete old log files.

kevlam2706 (approved):

nrgill28 (approved):

Issue #15, opened 2020-01-11
Prototyping: IR Distance Sensor


nicojopr:

kevlam2706:
Deliverable here is to create a wrapper class for the Sharp IR sensor that allows us to get either the raw voltage reading or a calculated distance value based on an equation which we can determine. Note that if we do provide a calculated distance, we need to test it and understand how accurate it is in real world conditions.

Citysurvivor:
Code is done, tested (IR sensor equation tested on a roborio), and builds successfully. Looks like I can go ahead and close this issue now.

PR #66, opened 2020-02-06
Created a class for an IR sensor.

Citysurvivor:
Class converts voltage from a short range IR sensor into distance. Returns fixed values when the distance exceeds or falls below predetermined values.

See ir sensor notes in the wiki for instructions, tips.

Issue #15

### DIFF
--- a/src/main/java/frc/robot/sensors/IrSensor.java
+++ b/src/main/java/frc/robot/sensors/IrSensor.java
@@ -6,7 +6,7 @@
 /*----------------------------------------------------------------------------*/
 
 /**
- * Code modified by Charlie MacDonald.
+ * Charlie MacDonald.
  * Last modified on Feb 8th, 2020.
  */
 
@@ -75,5 +75,6 @@ public class IrSensor implements Sendable {
 	@Override
 	public void initSendable(SendableBuilder builder) {
         builder.addDoubleProperty("Distance", this::getDistance, null);
+        builder.addDoubleProperty("Voltage", this::takeMeasurement, null);
 	}
 }
--- a/src/main/java/frc/robot/sensors/IrSensor.java
+++ b/src/main/java/frc/robot/sensors/IrSensor.java
@@ -38,10 +38,10 @@ public class IrSensor implements Sendable {
   
     public double takeMeasurement() {
         //Make sure to call this method in the periodic() method of the subsystem that uses the IR sensor. 
-        //This method continuously monitors the IR sensor output and filters outliers. Also returns sensor voltage, if anyone wishes to use that.
+        //This method continuously monitors the IR sensor output and filters outliers. Also returns sensor voltage if anyone wants to use that.
         
         irVoltage = (m_filter.calculate(m_irsensor.getValue())/1000); 
-        //Gets ir sensor value, filters outliers, and converts milivolt output to volts.
+        //Gets IR sensor value, filters outliers, and converts milivolt output to volts.
 
         return(irVoltage);
     }
--- a/src/main/java/frc/robot/sensors/IrSensor.java
+++ b/src/main/java/frc/robot/sensors/IrSensor.java
@@ -6,8 +6,8 @@
 /*----------------------------------------------------------------------------*/
 
 /**
- * Code by Charlie Macdonald.
- * Last modified on Feb 6th, 2020.
+ * Code modified by Charlie MacDonald.
+ * Last modified on Feb 8th, 2020.
  */
 
 package frc.robot.sensors;
@@ -25,21 +25,32 @@ public class IrSensor implements Sendable {
     private final AnalogInput m_irsensor;
 
     //Set the constants to calculate IR sensor distance:
-        //Values for short range (Sharp GP2Y0A41SK0F) 4-30cm sensor: 12.8528, 0.007545, -3.92, 30, 4.5, 0.
+        //Values for short range (Sharp GP2Y0A41SK0F) 4-30cm sensor: 12.8528, 0.007545, -3.92, 25, 4.5, 0.
         //Note: If a different IR sensor is used, these values will need to be updated.
     private final double IR_SLOPE_VALUE = 12.8528;
     private final double IR_X_VALUE = 0.007545;
     private final double IR_CONSTANT = -3.92;
     private final double IR_MAX_DISTANCE_CM = 25;
     private final double IR_MIN_DISTANCE_CM = 4.5;
-    private final double IR_MIN_OUTPUT_CM = 0;
+    private final double IR_MIN_OUTPUT_CM = 0.0;
+
+    private double irVoltage;
   
+    public double takeMeasurement() {
+        //Make sure to call this method in the periodic() method of the subsystem that uses the IR sensor. 
+        //This method continuously monitors the IR sensor output and filters outliers. Also returns sensor voltage, if anyone wishes to use that.
+        
+        irVoltage = (m_filter.calculate(m_irsensor.getValue())/1000); 
+        //Gets ir sensor value, filters outliers, and converts milivolt output to volts.
+
+        return(irVoltage);
+    }
+
     public double getDistance() {
         //Calculate distance (cm) from mV:
         //Accurate to +/- 2cm when measuring within 4.5-25cm.
         double currentDistanceCm = ((1/
-            (((m_filter.calculate(m_irsensor.getValue()))
-            /1000 - IR_X_VALUE)
+            ((irVoltage - IR_X_VALUE)
             /IR_SLOPE_VALUE)
             ) //Inverse is effectively calculated at this point.
             +IR_CONSTANT);
@@ -54,7 +65,6 @@ public class IrSensor implements Sendable {
         }
 
         return(currentDistanceCm);
-
     }
 
     public IrSensor(int channel) {
--- a/src/main/java/frc/robot/sensors/IrSensor.java
+++ b/src/main/java/frc/robot/sensors/IrSensor.java
@@ -1,0 +1,66 @@
+/*----------------------------------------------------------------------------*/
+/* Copyright (c) 2018-2019 FIRST. All Rights Reserved.                        */
+/* Open Source Software - may be modified and shared by FRC teams. The code   */
+/* must be accompanied by the FIRST BSD license file in the root directory of */
+/* the project.                                                               */
+/*----------------------------------------------------------------------------*/
+
+/**
+ * Code by Charlie Macdonald.
+ * Last modified on Feb 5th, 2020.
+ */
+
+package frc.robot.sensors;
+
+import edu.wpi.first.wpilibj.AnalogInput;
+import edu.wpi.first.wpilibj.MedianFilter;
+import edu.wpi.first.wpilibj.Sendable;
+import edu.wpi.first.wpilibj.smartdashboard.SendableBuilder;
+import edu.wpi.first.wpilibj.smartdashboard.SendableRegistry;
+
+public class IrSensor implements Sendable {
+    private static final int kIRPort = 1; //set IR sensor port
+        //TODO: The IR sensor's port may change, depending on the robot setup. (Use a robot-specific config to set this?)
+
+    //median filter to discard outliers. Filters over 10 samples
+    private final MedianFilter m_filter = new MedianFilter(10);
+
+    private final AnalogInput m_irsensor;
+
+    //Set the formula values to calculate IR sensor distance:
+        //Values for short range (5-30cm) sensor: 12.8528, -0.007545, -3.92, 35, 4.5.
+        //TODO: If a different IR sensor is used, these values will need to be updated. (Use a robot-specific config to set them?)
+    private final double irSlopeValue = 12.8528;
+    private final double irXValue = -0.007545;
+    private final double irConstant = -3.92;
+    private final double irMaxDistance = 35;
+    private final double irMinDistance = 4.5;
+  
+    public double getDistance() {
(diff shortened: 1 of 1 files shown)


src/main/java/frc/robot/sensors/IrSensor.java
```diff
@@ -0,0 +1,66 @@
...
+        //TODO: If a different IR sensor is used, these values will need to be updated. (Use a robot-specific config to set them?)
+    private final double irSlopeValue = 12.8528;
+    private final double irXValue = -0.007545;
+    private final double irConstant = -3.92;
+    private final double irMaxDistance = 35;
+    private final double irMinDistance = 4.5;
+  
+    public double getDistance() {
+        //Calculate distance (cm) from mV:
+        double currentDistance = ((1/(((m_filter.calculate(m_irsensor.getValue()))/1000 +irXValue)/irSlopeValue))+irConstant);
```
nrgill28: This line is pretty hard to read, maybe break it up a bit?
Also, the median filter might not be a good idea unless the sensor is constantly getting values.
For example: if the sensor is polled 20 times at 10cm, and then a minute later is polled once at 30cm the previous 9 polls are being included in this calculation when they are far off from the actual value.

src/main/java/frc/robot/sensors/IrSensor.java
```diff
@@ -0,0 +1,66 @@
...
+        //Accurate within +/- 1.5cm when measuring from 5-20cm.
+
+        //The following "if" statements fix the values outside the sensor's accurate measuring range.
+        if (currentDistance > irMaxDistance) {
+            currentDistance = 35; //set distance to a fixed value if it is above the max measuring range.
+        }
+        if (currentDistance < irMinDistance) {
+            currentDistance = 0; //set distance to a fixed value if it is below the min measuring range.
+            //Warning: Distances under 2cm will still produce erroneous values.
+        }
```
nrgill28, line 65: I think these two if statements can be replaced with `currentDistance = Math.min(Math.max(currentDistance, irMinDistance), irMaxDistance)`

(Effectively a clamp function)

src/main/java/frc/robot/sensors/IrSensor.java
```diff
@@ -0,0 +1,66 @@
...
+import edu.wpi.first.wpilibj.smartdashboard.SendableRegistry;
+
+public class IrSensor implements Sendable {
+    private static final int kIRPort = 1; //set IR sensor port
+        //TODO: The IR sensor's port may change, depending on the robot setup. (Use a robot-specific config to set this?)
+
+    //median filter to discard outliers. Filters over 10 samples
+    private final MedianFilter m_filter = new MedianFilter(10);
+
+    private final AnalogInput m_irsensor;
```
nrgill28, line 25: This one is a little nitpicky since it's about naming conventions, but I don't see the use for the `m_` prefix. I'd just remove that.

src/main/java/frc/robot/sensors/IrSensor.java
```diff
@@ -0,0 +1,66 @@
...
+
+    //Set the formula values to calculate IR sensor distance:
+        //Values for short range (5-30cm) sensor: 12.8528, -0.007545, -3.92, 35, 4.5.
+        //TODO: If a different IR sensor is used, these values will need to be updated. (Use a robot-specific config to set them?)
+    private final double irSlopeValue = 12.8528;
+    private final double irXValue = -0.007545;
+    private final double irConstant = -3.92;
+    private final double irMaxDistance = 35;
+    private final double irMinDistance = 4.5;
```
nrgill28: Are any of these tune-able? And if so, would it be beneficial to be able to edit them over NetworkTables without re-deploying the code? If so you may want to look at changing these into FluidConstants.

src/main/java/frc/robot/sensors/IrSensor.java
```diff
@@ -0,0 +1,66 @@
...
+
+import edu.wpi.first.wpilibj.AnalogInput;
+import edu.wpi.first.wpilibj.MedianFilter;
+import edu.wpi.first.wpilibj.Sendable;
+import edu.wpi.first.wpilibj.smartdashboard.SendableBuilder;
+import edu.wpi.first.wpilibj.smartdashboard.SendableRegistry;
+
+public class IrSensor implements Sendable {
+    private static final int kIRPort = 1; //set IR sensor port
+        //TODO: The IR sensor's port may change, depending on the robot setup. (Use a robot-specific config to set this?)
```
nrgill28: Yes, the channel for the sensor should be defined in the Config.java file with a robotSpecific value. This way it can automatically switch between channels based on which robot it's deployed to

nrgill28 (commented):
Great work! Code looks pretty good overall, just a few comments here and there on a couple things

Citysurvivor:
Note to self:

Some code edits were made in the time between the code being pushed and the code being reviewed. Check local files for changes.

src/main/java/frc/robot/sensors/IrSensor.java
kevlam2706: We could do two things..

1 - remove the median filter if we think this class will only be used for "ad hoc" readings

2 - add a "takeMeasurement()" method to the class and tell users to put a call to that in the periodic() method of whatever subsystem instantiates one of these. The method would then update the median filter.

src/main/java/frc/robot/sensors/IrSensor.java
kevlam2706: The m_ prefix is meant to indicate a member variable and was more popular as a formatting convention in my C++ days.

Yeah, if nobody else is using that convention then you can remove it.

src/main/java/frc/robot/sensors/IrSensor.java
kevlam2706: No, these are meant to curve-fit based on the physical response of the sensor. They shouldn't change.

One suggestion: maybe include the units? e.g. irMaxDistanceCm

src/main/java/frc/robot/sensors/IrSensor.java
kevlam2706: I think this should be a value passed into the constructor so I can do this:

```
IrSensor inputSensor = new IrSensor(0); // input sensor is connected to analog input 0
IrSensor outputSensor = new IrSensor(2); // output sensor is connected to analog input 2;
```

kevlam2706: I think this should be given in the constructor (see my other comment)

src/main/java/frc/robot/sensors/IrSensor.java
kevlam2706: Charlie can you write somewhere in the comments the exact model number of the sensor this class is meant to represent? Sharp model ABC123-whatever. Do you still have the data sheet on your computer? Otherwise I'll help you find that number.

kevlam2706 (approved):
Looks good!

src/main/java/frc/robot/sensors/IrSensor.java
Citysurvivor: I think I'll keep that naming convention for now since I'm more used to it.

src/main/java/frc/robot/sensors/IrSensor.java
Citysurvivor: Variables won't change often (only if we swap out IR sensors), so there wouldn't be much benefit from changing those into FluidConstants.

Added sensor model number, and added units to variable names.

kevlam2706 (approved):

src/main/java/frc/robot/sensors/IrSensor.java
```diff
@@ -0,0 +1,71 @@
...
+
+    //median filter to discard outliers. Filters over 10 samples
+    private final MedianFilter m_filter = new MedianFilter(10);
+
+    private final AnalogInput m_irsensor;
+
+    //Set the formula values to calculate IR sensor distance:
+        //Values for short range (Sharp GP2Y0A41SK0F) 4-30cm sensor: 12.8528, -0.007545, -3.92, 30, 4.5.
+        //Note: If a different IR sensor is used, these values will need to be updated.
+    private final double irSlopeValue = 12.8528;
```
georgetzavelas: Since these are constants do you mind CAPITALIZING them?

src/main/java/frc/robot/sensors/IrSensor.java
```diff
@@ -0,0 +1,71 @@
...
+        double currentDistanceCm = ((1/
+            (((m_filter.calculate(m_irsensor.getValue()))
+            /1000 - irXValue)
+            /irSlopeValue)
+            ) //Inverse is effectively calculated at this point.
+            +irConstant);
+
+        //The following "if" statements fix the values outside the sensor's accurate measuring range.
+        if (currentDistanceCm > irMaxDistanceCm) {
+            currentDistanceCm = 30; //set distance to a fixed value if it is above the max measuring range.
```
georgetzavelas: Can we make this number to the const you defined above?

src/main/java/frc/robot/sensors/IrSensor.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+        if (currentDistanceCm < irMinDistanceCm) {
+            currentDistanceCm = 0; //set distance to a fixed value if it is below the min measuring range.
```
georgetzavelas: Can you define a new const for this? It will make the code more readable.

src/main/java/frc/robot/sensors/IrSensor.java
```diff
@@ -0,0 +1,71 @@
...
+            //Warning: Distances under 1cm will still produce erroneous values.
+        }
+
+        return(currentDistanceCm);
+
+    }
+
+    public IrSensor(int channel) {
+        this.m_irsensor = new AnalogInput(kIRPort);
```
georgetzavelas: Channel is being passed in here but I don't see it being used.

georgetzavelas (commented):
Great job Charlie! Some comments for your consideration.

src/main/java/frc/robot/sensors/IrSensor.java
Citysurvivor: Sure thing. Also, I believe words in constants are separated by underscores, i.e. IR_SLOPE_VALUE, so I'll do that.

src/main/java/frc/robot/sensors/IrSensor.java
Citysurvivor: > Can we make this number to the const you defined above?

OK, did that for max distance. 

As for minimum distance, I think the code should set currentDistanceCm to 0 if the measured value falls below the minimum distance. The reasoning being that, if an object was inside the blindspot between the sensor and the sensor's minimum range, it would be safer to set the variable to zero (and assume the object is touching) than to assume the object is located around the minimum range.

src/main/java/frc/robot/sensors/IrSensor.java
Citysurvivor: @georgetzavelas Not sure if I understand. Should I use a variable in place of the zero? I think currentDistanceCm should always be set to zero if the measured distance is below the sensor's minimum range.

Citysurvivor: OK, used a constant in place of zero.

src/main/java/frc/robot/sensors/IrSensor.java
Citysurvivor: Actually, I think I will keep the if statement. If the measured distance falls below the minimum value (4.5), I will set it to a different constant (0) instead of setting it to match the minimum value (4.5). I'm not sure if that can be done with that math.min/math.max code, so I'll stick with an if statement.

src/main/java/frc/robot/sensors/IrSensor.java
Citysurvivor: Whoops, fixed that. Channel will be used to set the IR sensor channel instead of having it hardcoded.

src/main/java/frc/robot/sensors/IrSensor.java
Citysurvivor: Yup, fixed that. Channel number can now be passed as an argument.

src/main/java/frc/robot/sensors/IrSensor.java
Citysurvivor: Added a takeMeasurement() method to be called in a periodic method. Updates the median filter and returns the voltage, if anyone wishes to use the latter.

src/main/java/frc/robot/sensors/IrSensor.java
```diff
@@ -0,0 +1,80 @@
+/*----------------------------------------------------------------------------*/
+/* Copyright (c) 2018-2019 FIRST. All Rights Reserved.                        */
+/* Open Source Software - may be modified and shared by FRC teams. The code   */
+/* must be accompanied by the FIRST BSD license file in the root directory of */
+/* the project.                                                               */
+/*----------------------------------------------------------------------------*/
+
+/**
```
georgetzavelas, line 8: Nice touch but this is not needed since git has a version history of who modified what and when. Also it is misleading since someone else could come and change a few lines in this file.

georgetzavelas (approved):
Great refactor!

src/main/java/frc/robot/sensors/IrSensor.java
Citysurvivor: OK, will keep in mind for future commits. Won't change it for now since it would require the merge request to be re-approved, and the issue is fairly minor .

Issue #34, opened 2020-01-18
Feeder/Indexer Subsystem


nicojopr:
The feeder/indexer provides power cell storage (4 cells in hopper and one in funnel) and the ability to feed the shooter. It will have one motor with encoder.

Production software for the feeder/indexer subsystem in the robot code, including:

- interface definitions, motors, sensors
- commands
- etc..
- tested independently
- merged and integrated with 2020 robot code in github
- ready for integration

nicojopr:
@kevlam2706 comment from duplicate :
I just noticed that the Feeder wheel may be separate from the Indexer wheel. In the prototype the two were tied together by a belt. I’ve asked on Slack to clarify if the real one will be separately controlled, in which case we will need to add a second motor to this subsystem and include in the commands the logic to manage the timing and sequencing between them.

kevlam2706:
Some terminology update since I keep getting this wrong...

The big wheel inside the hopper is the **Feeder**. Its job is to Feed the shooter.

The little wheel at the edge of the funnel is the **Indexer**. Its job is to ensure the balls enter the hopper with a small separation between each ball so they don't jam up.

The Feeder and Indexer wheels will both be driven by the same motor on a single belt. The Indexer wheel is designed to rotate slower than the Feeder wheel so that each ball ends up spaced out a bit when inside the hopper.

There will be an IR sensor to detect when a ball is sitting at the Indexer. This should trigger the Feeder to "increment" and move everything closer to the shooter by a predetermined distance.

We should be able to keep track of how many balls are in the hopper. This will be used when shooting and also when intaking. If we already have four balls in the hopper and one at the indexer, for example, then we may want to disable the intake so that we don't take any more.

AbhiJ2706:
@kevlam2706 I believe we will need another IR sensor at the shooter end of the indexer track in order to monitor the amount of balls in the indexer. This will count the number of outgoing balls, so that that number can be subtracted from the count the first IR sensor creates (which would add 1 for each incoming ball). That way, the exact number of balls in the indexer will known.

kevlam2706:
We should check with controls and D&F to see if they are planning on this.

PR #67, opened 2020-02-06
Feeder subsystem

AbhiJ2706:
Adding the code for the feeder subsystem. This contains the feeder subsystem code, commands to run the feeder and increment the feeder, and mapped these commands to joystick buttons. 

Further changes will need to be made to configure it for the 2020 robot, and once this code has been tested.

Issue #34


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -64,6 +67,12 @@ private void configureButtonBindings() {
     // Instantiate the intake command and bind it
     intakeCommand = new OperatorIntakeCommand();
     new JoystickButton(driverStick, XboxController.Button.kBumperLeft.value).whenHeld(intakeCommand);
+
+    incrementFeederCommand = new IncrementFeeder();
+    new JoystickButton(controlStick, XboxController.Button.kA.value).whenPressed(incrementFeederCommand);
+
+    emptyFeederCommand = new IncrementFeeder();
```
georgetzavelas: Do you want this to reference `EmptyFeeder`?

src/main/java/frc/robot/config/FluidConstant.java
```diff
@@ -76,14 +76,6 @@ public T getValue() {
...
     }
     
-    /**
-     * Get method for implementing the Supplier interface
-     * @return The current value of this constant
-     */
-    public T get() {
-        return this.value;
-    }
-
```
kevlam2706: Was this block deleted intentionally?

src/main/java/frc/robot/subsystems/FeederSubsystem.java
```diff
@@ -0,0 +1,130 @@
...
+
+    //The motor that drives the feeder
+    private static WPI_TalonSRX feederTalon;
+
+    //IR sensor that monitors the indexer
+    private static AnalogInput indexerIrSensor;
+
+    //How much to shift the feeder wheel when incrementing
+    public static FluidConstant<Double> INCREMENT_TICKS = new FluidConstant<>("IncrementTicks", 0.0)
+                .registerToTable(Config.constantsTable);
```
kevlam2706, line 36: maybe name this with a FeederSubsystem prefix so it is clear what it is?

src/main/java/frc/robot/subsystems/FeederSubsystem.java
```diff
@@ -0,0 +1,130 @@
...
+    //Max distance at which the robot knows a ball is at the indexer
+    public static FluidConstant<Integer> IR_MAX_DISTANCE = new FluidConstant<>("IrMaxDistance", 0)
```
kevlam2706: maybe name with a FeederSubsystem prefix so it's clear what it is?

src/main/java/frc/robot/subsystems/FeederSubsystem.java
```diff
@@ -0,0 +1,130 @@
...
+        feederTalon.configPeakOutputReverse(-FEEDERSUBSYSTEM_PEAK_OUTPUT.get(), kTimeoutMs);
+        feederTalon.configAllowableClosedloopError(0, 0, kTimeoutMs);
+        feederTalon.config_kF(kPIDLoopIdx, FEEDERSUBSYSTEM_F.get(), kTimeoutMs);
+        feederTalon.config_kP(kPIDLoopIdx, FEEDERSUBSYSTEM_P.get(), kTimeoutMs);
+        feederTalon.config_kI(kPIDLoopIdx, FEEDERSUBSYSTEM_I.get(), kTimeoutMs);
+        feederTalon.config_kD(kPIDLoopIdx, FEEDERSUBSYSTEM_D.get(), kTimeoutMs);
+        int absolutePosition = feederTalon.getSensorCollection().getPulseWidthPosition();
+        feederTalon.setSelectedSensorPosition(absolutePosition, kPIDLoopIdx, kTimeoutMs);
+        feederTalon.setSelectedSensorPosition(0, 0, 10);
+
```
kevlam2706, line 84: Some of these calls return errorCodes - most critically the call to configSelectedFeedbackSensor(). We need to catch the error code and do something about anything abnormal, such as if the encoder isn't found by the Talon.

Nathan has written a cool conditions framework that lets you set a condition called "encoderFound" or "encoderHealthy" or whatever you want to call it. The condition framework allows the subsystem to determine if it should run or not based on every condition being met. You could use that to essentially flag the subsystem as being unable to run if the errorCode is not happy. We should also log the errors.

kevlam2706 (approved):
Looks good. Look into incorporating the Subsystem Conditions framework that Nathan set up as I think it would be a good way to catch error states and prevent the system from doing anything if, say, the encoder is disconnected or not found.

src/main/java/frc/robot/config/FluidConstant.java
georgetzavelas: This block seems to be deleted in many of the open PRs so I suspect it may be occurring automatically by the IDE as it is possibly unused?

AbhiJ2706: When I opened the file there were 2 of that method and VSCode was giving an error.

LaraLim (approved):
Looks good.

Zinka010 (approved):

PR #68, opened 2020-02-08
Shooter Subsystem - Issue #45

### LINKED ISSUE #20: Prototyping: Operator Control Rotate/Aim Robot

### LINKED ISSUE #45: Shooter Subsystem
Production software for the shooter subsystem in the robot code, including:
- interface definitions
- commands
- PID
- etc..
- tested independently
- merged and integrated with 2020 robot code in github
- ready for integration

### LINKED ISSUE #88: Refactor shooter subsystem to use NEO motor
D&F would like to switch from the 2x775 motor assembly to use a single Rev Robotics NEO brushless motor controlled by a Spark Max motor controller.

This will require the shooter subsystem to be modified. The commands and general structure don't need to change. The good news is that the Spark Max has a pretty good software API, examples on GitHub, and control of a single motor with a built-in encoder is quite easy. (Possibly even easier than what we were going to do with the 2x775 approach)

It's also quite comparable to the Talon SRX code from CTRE, so it should look fairly familiar.

See here for example code:

https://github.com/REVrobotics/SPARK-MAX-Examples/tree/master/Java/Velocity%20Closed%20Loop%20Control

You will also need to download the Spark Max libraries to your computer, change the build.gradle to point to those new vendor libraries.

To test, we will either get a standalone NEO and Spark Max to connect to one of the robots, or we will switch out the existing CIM motor in the prototype fuel shooter.

LaraLim:
ShooterSubsystem, Config, and the SpinUpShooter command currently use some protobot-specific values, like the CAN ID constant and PID values. This pull request will not be merged until the code is properly tested on protobot. Parts of the code that are relevant only to protobot will then be changed or removed.


georgetzavelas:
Looks like you need a rebase

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -64,6 +66,10 @@ private void configureButtonBindings() {
     // Instantiate the intake command and bind it
     intakeCommand = new OperatorIntakeCommand();
     new JoystickButton(driverStick, XboxController.Button.kBumperLeft.value).whenHeld(intakeCommand);
+
+    // Instantiate the shooter ramping command and bind it
+    rampShooterCommand = new SpinUpShooter();
+    new JoystickButton(driverStick, XboxController.Button.kA.value).whenPressed(rampShooterCommand);
```
kevlam2706: How will this work - will it hold the shooter flywheel at speed while the button is pressed down?

src/main/java/frc/robot/commands/SpinUpShooter.java
```diff
@@ -0,0 +1,37 @@
...
+    addRequirements(shooterSubsystem);
+  }
+
+  @Override
+  public void initialize() {
+  }
+
+  @Override
+  public void execute() {
+    // Hardcoded RPM for testing
```
kevlam2706: maybe set a variable in the command that we temporarily set to 2000 RPM for testing, but eventually will be determined by a call to some kind of formula. Then refer to the variable on line 22 and 24.

src/main/java/frc/robot/config/FluidConstant.java
```diff
@@ -76,14 +76,6 @@ public T getValue() {
         return this.value;
     }
     
-    /**
-     * Get method for implementing the Supplier interface
-     * @return The current value of this constant
-     */
-    public T get() {
-        return this.value;
```
kevlam2706: I notice a lot of people who have this deletion in their code reviews. Any idea what's going on?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,107 @@
...
+  // Singleton instance for ShooterSubsystem
+  private static final ShooterSubsystem INSTANCE_SHOOTER = new ShooterSubsystem();
+
+  private WPI_TalonSRX m_shooter;
+
+  // Indicates use of the primary PID loop rather than cascaded ones
+  int kPIDLoopIDX = 0;
+
+	 // Set to zero to skip waiting for confirmation, set to nonzero to wait and
+	 // report to DS if action fails.
```
kevlam2706: watch out for your indenting..

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,107 @@
...
+    if (Config.SHOOTER_MOTOR != -1){
+      m_shooter = new WPI_TalonSRX(Config.SHOOTER_MOTOR);
+    }
+
+    // Factory Default to prevent unexpected behaviour
+    m_shooter.configFactoryDefault();
+
+    // Config sensor used for PID control
+    m_shooter.configSelectedFeedbackSensor(FeedbackDevice.CTRE_MagEncoder_Relative, 
+    kPIDLoopIDX, kTimeoutMs);
```
kevlam2706: eventually add error checking here in case the encoder isn't connected, we don't want to try running a closed loop mode

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,107 @@
...
+    m_shooter.configFactoryDefault();
+
+    // Config sensor used for PID control
+    m_shooter.configSelectedFeedbackSensor(FeedbackDevice.CTRE_MagEncoder_Relative, 
+    kPIDLoopIDX, kTimeoutMs); 
+
+    // Invert the phase of the sensor so positive output yields positive change
+    m_shooter.setSensorPhase(true);
+    
+		// Config the peak and nominal outputs
```
kevlam2706: indenting

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,107 @@
...
+      return INSTANCE_SHOOTER;
+  }
+
+  /**
+   * Return the motor velocity measured by the encoder
+   */
+  public double getVelocity(){
+    double encoderVelocity = m_shooter.getSelectedSensorVelocity();
+    // Add the ability to print to SmartDashboard?
+    return encoderVelocity;
```
kevlam2706: what are the units here? can we turn them back into RPM?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,107 @@
...
+  public void checkRPM(int targetRPM){
+
+    // Calculate RPM based on the encoder reading
+    double calculatedRPM = (m_shooter.getSelectedSensorVelocity() * 600) / 4096;
+
+    // Verify that the motor is running at the target RPM
+    if (calculatedRPM < (targetRPM + 50) && calculatedRPM > (targetRPM - 50)){
+      System.out.println("calculatedRPM is within 50 units of targetRPM");
+      // I have no idea if +/- 50 tolerance around the RPM is accurate enough
+      // Test and change, add setting conditions
```
kevlam2706, line 96: Suggest that we put the tolerance value (50) in a class variable so we can easily change it. Then just refer to it here.

Also, this method should be public boolean instead of public void, so that it can return true if at speed or false if not.

kevlam2706 (approved):
Looks great! Some minor formatting fixes and other suggestions.

src/main/java/frc/robot/commands/SpinUpShooter.java
```diff
@@ -0,0 +1,37 @@
...
+    shooterSubsystem.setRPM(2000);
```
georgetzavelas: Can we make the number into a constant?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,107 @@
...
+
+import com.ctre.phoenix.motorcontrol.ControlMode;
+import com.ctre.phoenix.motorcontrol.FeedbackDevice;
+import com.ctre.phoenix.motorcontrol.can.WPI_TalonSRX;
+import frc.robot.config.Config;
+
+public class ShooterSubsystem extends SubsystemBase {
+
+  // Singleton instance for ShooterSubsystem
+  private static final ShooterSubsystem INSTANCE_SHOOTER = new ShooterSubsystem();
```
georgetzavelas: Instead of initializing here, consider using a Singleton pattern that lazy loads instead of relying on static loading like this: https://github.com/georgetzavelas/guava-examples/blob/master/src/com/design/pattern/creational/singleton/Singleton.java

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,107 @@
...
+    // Add the ability to print to SmartDashboard?
+    return encoderVelocity;
+  }
+
+  /**
+   * Set the target RPM to ramp up to. Will eventually get this value 
+   * based on distance from power port or other factors.
+   */
+  public void setRPM(int targetRPM){
+    velocityModeUnits = targetRPM * 4096 / 600;
```
georgetzavelas: Can we make these numbers into constants?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,107 @@
...
+   * Check the actual RPM and compare it with targetRPM
+   * to verify that the shooter is up to necessary speed to fire.
+   */
+  public void checkRPM(int targetRPM){
+
+    // Calculate RPM based on the encoder reading
+    double calculatedRPM = (m_shooter.getSelectedSensorVelocity() * 600) / 4096;
+
+    // Verify that the motor is running at the target RPM
+    if (calculatedRPM < (targetRPM + 50) && calculatedRPM > (targetRPM - 50)){
```
georgetzavelas: Can we make these numbers into constants?

georgetzavelas (commented):
Awesome progress! Some comments for your consideration

src/main/java/frc/robot/RobotContainer.java
LaraLim: The command will be scheduled when the button goes from inactive to active. The button doesn't have to be held continuously for the shooter to ramp up. The command can be rebound to another trigger based on drive team preference or anything that makes more sense.

src/main/java/frc/robot/config/FluidConstant.java
nrgill28: This issue is fixed in master. Can you revert the changes to this file so we don't break it again?

nrgill28 (changes requested):

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
LaraLim: The units read off the encoder are position change per 100ms. CheckRPM() uses it to calculate the current RPM when the shooter is running and compares it with targetRPM.

src/main/java/frc/robot/config/FluidConstant.java
georgetzavelas: Rebasing should simply remove this problem, no need to explicitly revert anything.

Zinka010 (approved):
Seems good for now considering the changes

src/main/java/frc/robot/subsystems/DriveBase.java
```diff
@@ -54,7 +54,7 @@ private DriveBase() {
         talon5plyboy = new WPI_TalonSRX(Config.TALON_5_PLYBOY);
 
         robotDriveBase = new DifferentialDrive(leftFrontTalon, rightFrontTalon);
-
+        _pidgey = new PigeonIMU (Config.robotSpecific(null, null, null, leftFrontTalon, leftRearTalon, talon5plyboy));
```
georgetzavelas: Is this needed as I don't see it being used anywhere?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,95 @@
...
+import com.revrobotics.ControlType;
+import com.revrobotics.CANSparkMaxLowLevel.MotorType;
+
+public class ShooterSubsystem extends SubsystemBase {
+
+  private CANSparkMax m_shooter;
+  private CANPIDController m_pidController;
+  private CANEncoder m_encoder;
+
+  double kF = 0.00018;
```
georgetzavelas: Can you make these constants? Also what are the values for, e.g. kF == SHOOTER_FUDGE_FACTOR?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,95 @@
...
+   */
+  public void setRPM(int inputRPM){
+    setpointRPM = inputRPM;
+  }
+
+  /**
+   * Return the motor velocity measured by the encoder
+   */
+  public double getRPM(){
+    double encoderRPM = m_encoder.getVelocity();
```
georgetzavelas: nitpick: you don't need the temporary variable here.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,95 @@
...
+   * Returns the singleton instance for the ShooterSubsystem
+   */
+  public static ShooterSubsystem getInstance() {
+    return ShooterHolder.INSTANCE_SHOOTER;
+  }
+
+  /**
+   * Set the target RPM to ramp up to.
+   */
+  public void setRPM(int inputRPM){
```
georgetzavelas: How about `setTargetRPM`?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,95 @@
...
+  public double getRPM(){
+    double encoderRPM = m_encoder.getVelocity();
+    return encoderRPM;
+  }
+
+  /**
+   * Check the actual RPM and compare it with targetRPM
+   * to verify that the shooter is up to necessary speed to fire.
+   */
+  public boolean checkRPM(int inputRPM){
```
georgetzavelas: Do you expect `inputRPM` to be different than what is currently stored for `setpointRPM`? If not then it would may make sense to simply use that without needing to pass in a variable. Also this function feels more like a `isAtTargetRPM` method

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,95 @@
...
+    double encoderRPM = m_encoder.getVelocity();
+    return (encoderRPM < (inputRPM + RPM_TOLERANCE) && encoderRPM > (inputRPM - RPM_TOLERANCE));
```
georgetzavelas: How about
```
return (Math.abs(inputRPM - encoderRPM) < RPM_TOLERANCE)
```

georgetzavelas (commented):
This is coming along great Lara! I left comments in one Subsystem that applies to the other.

src/main/java/frc/robot/subsystems/DriveBase.java
Lockscout: This is code that I have written that ended up on her branch. The code here is outdated and if it were to be put into master, it would break master. Currently communicating to fix it. But yeah this code is needed, as its used in drivetrainPID stuff.

Lockscout: Ok it isnt used, like you said. I misunderstood the issue. That line just needs to be deleted. Sorry.

nrgill28 (approved):

georgetzavelas (approved):

Lockscout:
Alright, so here's what we've concluded after testing today.

The shooter ramping, and pid works great! Of course it still needs a bit of tuning, but overall it works good.

There is one issue with it. After you turn on the robot, or upload new code to the roboRIO, The target speed is always set to 2000 (as that's what the default is) and causes it to ramp up to that rpm when its enabled, until you hit the button once to set the target to 0. What might need to be done is when the robot is initializing, set the target to 0.

PR #82, opened 2020-02-11
Added Bling Subsystem #27

### LINKED ISSUE #27: Bling Subsystem
Controlling colours for:
- fun
- driver feedback

Involves writing to network tables from roboRio to communicate with raspberry Pi

Definition of Done:
Production software for the bling subsystem in the robot code, including:
- interface definitions
- commands
- etc..
- tested independently
- merged and integrated with 2020 robot code in github
- ready for integration

jwt2706:
It's just the frame work.

### DIFF
--- a/src/main/java/frc/robot/subsystems/Bling.java
+++ b/src/main/java/frc/robot/subsystems/Bling.java
@@ -1,0 +1,35 @@
+/*----------------------------------------------------------------------------*/
+/* Copyright (c) 2019 FIRST. All Rights Reserved.                             */
+/* Open Source Software - may be modified and shared by FRC teams. The code   */
+/* must be accompanied by the FIRST BSD license file in the root directory of */
+/* the project.                                                               */
+/*----------------------------------------------------------------------------*/
+
+package frc.robot.subsystems;
+
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+
+public class Bling extends SubsystemBase {
+
+
+  private static Bling INSTANCE = new Bling();
+  /**
+   * Creates a new Bling.
+   */
+  private Bling() {
+
+
+  }
+
+  public static Bling getINSTANCE() {
+    return INSTANCE;
+  }
+
+  @Override
+  public void periodic() {
+    // This method will be called once per scheduler run
+  
+
+    
+  }
+}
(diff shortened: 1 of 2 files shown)


src/main/java/frc/robot/subsystems/Bling.java
```diff
@@ -0,0 +1,35 @@
...
+/*----------------------------------------------------------------------------*/
+
+package frc.robot.subsystems;
+
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+
+public class Bling extends SubsystemBase {
+
+
+  private static Bling INSTANCE = new Bling();
```
nicojopr, line 15: Suggest you follow the Singleton pattern and move the instantiation ("new Bling()") inside the getInstance or constructor

src/main/java/frc/robot/subsystems/Bling.java
```diff
@@ -0,0 +1,35 @@
...
+  /**
+   * Creates a new Bling.
+   */
+  private Bling() {
+
+
+  }
+
+  public static Bling getINSTANCE() {
```
nicojopr, line 24: Typically the getINSTANCE will check for null INSTANCE (and create the instance in this method if null) before returning INSTANCE

kevlam2706 (approved):
🤷🏻‍♂️😂

Not much to review! But maybe good to get the framework into master quickly so you can work on it more later.

Update: oops maybe I read too quickly. Please address Johanna’s comments

nicojopr (changes requested):
At least the null pointer check in getINSTANCE should be addressed before merging

src/main/java/frc/robot/subsystems/Bling.java
jwt2706: I originally did that, but when I got Nathan to have a look at it, he said that I could move it there.

src/main/java/frc/robot/subsystems/Bling.java
jwt2706: Oh, I see... I haven't done that for shuffleboard either, should I change it on both?

nicojopr (approved):
Discussed the getINSTANCE risk of non-null INSTANCE, will accept as is for singleton

PR #83, opened 2020-02-11
Pigeon Bullet Proofing

Lockscout:


nrgill28 (approved):

LaraLim (approved):

src/main/java/frc/robot/subsystems/DriveBase.java
```diff
@@ -51,10 +51,16 @@ private DriveBase() {
...
+        talon5plyboy = new WPI_TalonSRX((Config.TALON_5_PLYBOY));
+
         robotDriveBase = new DifferentialDrive(leftFrontTalon, rightFrontTalon);
-        _pidgey = new PigeonIMU (Config.robotSpecific(null, null, null, leftFrontTalon, leftRearTalon));
 
-        _pidgey.setFusedHeading(0.0, 30);
+
+        var pigeonTalon = Config.robotSpecific(null, null, rightRearTalon, leftFrontTalon, leftRearTalon, talon5plyboy);
+        if(pigeonTalon != null){
+            _pidgey = new PigeonIMU (pigeonTalon);
```
nicojopr, line 61: do we need to check _pidgey is non-null as well before calling setFusedHeading?

src/main/java/frc/robot/commands/DrivetrainPIDTurnDelta.java
```diff
@@ -58,6 +58,14 @@ public DrivetrainPIDTurnDelta(double deltaDegree, double forwardSpeed) {
         double currentAngle = drivebase.getCurrentAngle();
         logger.addHandler(Config.logFileHandler);
 
+        if(_pidgey == null){
+
```
nicojopr, line 62: What is best practice in the robot code if components are unavailable? Log it at least in my opinion.

src/main/java/frc/robot/subsystems/DriveBase.java
Lockscout: From what we have tested, its not needed to detect that for the setFusedHeading

src/main/java/frc/robot/commands/DrivetrainPIDTurnDelta.java
Lockscout: Thats true, Ill look into it

PR #84, opened 2020-02-11
An assorted collection of various improvements and refactors

nrgill28:
I decided to go through the project and clean some stuff up. I'll make these PRs in portions so it's not an overwhelming 20-file review for people.


kevlam2706 (approved):
OK. I suggest that for cosmetic things, try to get the original authors to follow the standard that you are trying to set. (And obviously whenever you add code to the repo, make sure it's very easy to follow and emulate). People, particularly rookies, may feel bad if they perceive that their code is getting "fixed" by others.

src/main/java/frc/robot/subsystems/DriveBase.java
```diff
@@ -20,14 +20,12 @@
 public class DriveBase extends SubsystemBase {
 
     // DriveBase is a singleton class as it represents a physical subsystem
-    private static DriveBase currentInstance;
+    private static DriveBase INSTANCE = new DriveBase();
```
georgetzavelas: Can we revert this to the old Singleton Pattern? Or better yet this pattern: https://github.com/georgetzavelas/guava-examples/blob/master/src/com/design/pattern/creational/singleton/Singleton.java

georgetzavelas (commented):
Any reason to remove the Xbox class?

nrgill28:
The Xbox class was removed since it's a duplicate of a WPILib built-in.
Also, the DriveBase singleton pattern matches the one you've linked (The only difference I see is the lack of the `final` modifier)

georgetzavelas:
I see about the XBox class. The singleton pattern is different than the one I linked. It uses an inner static class not a static variable.

nrgill28:
Oh. I don't know how I missed that...
I had updated the singleton pattern to be consistent with the rest of the subsystems in the code, should we switch to using this new one for all future singletons then?

georgetzavelas:
I would prefer us to use the inner static class way as it is thread safe. The only caveat with this approach is that initialization is for a failure state.

nicojopr (changes requested):
Will need to rebase again and retest

nrgill28:
This has become a little too outdated to merge easily. I'll just close this for now as it's not necessary.

PR #86, opened 2020-02-11
Added robot halt function

### LINKED ISSUE #77: Graceful failure if Robot Configuration is not found
As we discovered on Saturday, it is possible that the robot specific constants can fail, if the robot ID cannot be found.

The original default action was to assume a robot ID (0). However, this causes problems with some robots. Motors drive into each other, etc. This is not good.

Add some bulletproofing to the Constants system and to the whole robot initialization. If the robot ID cannot be found - do not do anything. We don't know if the robot has Talons or Victors, we don't know if they're going to drive against each other, we don't know if there's an intake or an arm or not... anything we default to could potentially cause a hardware failure or a software exception. So.... we should do nothing.

If the robot ID cannot be found (which should never happen after we've set it up), log it as a severe error, and don't allow Robot.java or any of its subsystems to initialize anything. The drivetrain should not be set up, driving commands and buttons should do nothing except possibly stream error messages to the log files.

nrgill28:
Oh and I reformatted one of the files to use 4-space indents.
Closes #77

### DIFF
--- a/src/main/java/frc/robot/Robot.java
+++ b/src/main/java/frc/robot/Robot.java
@@ -178,19 +178,28 @@ public class Robot extends TimedRobot {
      * Halts the robot in every way
      */
     public static void haltRobot(String s) {
-        // Disable the scheduler and cancel everything
+        /*
+         Disable the scheduler and cancel everything.
+         This stops all subsystems and commands and prevents new ones from starting
+        */
         var scheduler = CommandScheduler.getInstance();
         scheduler.disable();
         scheduler.cancelAll();
         
-        // Log the message
+        /*
+         Log the message and close the log file.
+         Not strictly necessary, but it's good practice to close files when you're done with it.
+        */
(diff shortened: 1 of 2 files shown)


kevlam2706 (approved):
Cool.

src/main/java/frc/robot/Robot.java
```diff
@@ -24,148 +27,184 @@
...
+        scheduler.cancelAll();
+        
+        // Log the message
+        logger.severe("ROBOT HALTED: " + s);
+        
+        // Close the log file
+        Config.logFileHandler.close();
+        
+        // Hang
+        while(true){}
```
georgetzavelas: Why would we do this? Can you add more comments / rationale. At face value this looks wrong but I understand you may be doing this for a specific reason. e.g. why not simply `System.exit(0);`

georgetzavelas (approved):

PR #107, opened 2020-02-20
Refactored drive base classes to allow for multiple drivebases

nrgill28:
basically, this allows for us to keep the existing drive base working properly while still being able to update it with the new drive base, since this upgrade will break the existing robots.

Notable changes:
 - Drive bases now extend from the `DriveBase` class
 - The new drive base class implements all shared methods and provides abstract methods to implement which will be specific to that drive base.
 - Made a lot of the drive base code more consistent and well documented.

It's a fairly large PR, I've tested it on the robots and everything works the same as it did before.


nrgill28:
Fixed all merge issues and re-tested on the robot

georgetzavelas (approved):

Lockscout (approved):
👍

PR #108, opened 2020-02-22
Updated to 2020 connection map

Zinka010:
- Updated to 2020 connection map
 - Updated to the new WPILib Button

**Note that on robots without a Victor SPX on CAN ID 6, this code will give continuous error messages but still work. This is the case in the master branch too**


Lockscout (approved):
Looks good, new connections seem ok.

georgetzavelas (approved):
😍

PR #110, opened 2020-02-22
Issue #45 - Shooter subsystem improvements

LaraLim:
Added a few simple methods (to return motor temperature and current draw), and a temporary fix for robots without the SPARK/NEO shooter attached.


Lockscout (commented):
looks good for a quick fix. Will we want it in the future, or should it be removed entirely?

georgetzavelas (approved):
shufflin-robot

Lockscout (approved):
Looks good. Thanks for adding the temp and current functions.

Issue #13, opened 2020-01-11
Prototyping: Current Limiting on Drivetrain


nicojopr:
-take an existing robot and apply current limiting profiles to the drivetrain
-goal is to shut down the motors if they're stalling for too long (pushing against opponent) (or apply motor brake?)
-but do not prevent ability to fight back minor skirmishes or recover from crashing into a field obstacle
-add a warning system before motors cut off so that drive team does not get surprised by a sudden power loss?

kevlam2706:
Use Mergonaut for this task as we already know how it behaves without the current limiting profiles.

Citysurvivor:
Starting now on Feb 10th. Open to collab with other students if anyone wants to.

kevlam2706:
Suggest the following resources to help:

https://phoenix-documentation.readthedocs.io/en/latest/ch13_MC.html

Feel free to Google and search places like the Chief Delphi forums to see what other teams are doing for drivetrain current limiting, though I recognize that this is like trying to drink from a firehose.

PR #114, opened 2020-02-25
Motor Current Limiter (Issue #13)

Citysurvivor:
Code limits motor current draw to 40 amps (continuously). Current that exceeds 60 amps for more than a second will be clamped down to 40. Code sends motor current info to Shuffleboard. Code can trigger driver feedback in teleopPeriodic if desired.

The changes to files other than Robot.java, Drivebase.java, and Config.java occurred when I updated my local vendor libraries - let me know if this isn't OK.

### DIFF
--- a/src/main/java/frc/robot/Robot.java
+++ b/src/main/java/frc/robot/Robot.java
@@ -40,8 +40,6 @@ public class Robot extends TimedRobot {
     //flag to indicate the real match
     private Boolean bRealMatch;
 
-    //create variable for current monitoring
-    double motorCurrent;
 
     /**
      * Determines if the robot is in a real match.
@@ -144,9 +142,9 @@ public class Robot extends TimedRobot {
         SmartDashboard.putNumber("Pigeon Yaw", DriveBase.getInstance().getCurrentAngle());
         CommandScheduler.getInstance().run();
 
-        //Send motor current levels to shuffleboard.
-        motorCurrent = DriveBase.leftFrontTalon.getSupplyCurrent();
-		SmartDashboard.putNumber("Current (Front Left)", motorCurrent);
+        //The following 2 methods run Drivebase methods that tell shuffleboard what the motor current draw is and if motor current limiting is active.
+        DriveBase.getMotorCurrent();
+        DriveBase.isMotorLimitActive();
     }
 
     /**
@@ -200,6 +198,12 @@ public class Robot extends TimedRobot {
      */
     @Override
     public void teleopPeriodic() {
+        //If motor current limiting is active, trigger driver feedback (notify driver that drivetrain power is reduced).
+        if (DriveBase.isMotorLimitActive() == true) {
+            //TODO: Put driver feedback code here. Consult with drive team's preferences.
+        } else {
+            //Disable current limit driver feedback here.
+        }
     }
 
     @Override
--- a/src/main/java/frc/robot/Robot.java
+++ b/src/main/java/frc/robot/Robot.java
@@ -141,9 +141,9 @@ public class Robot extends TimedRobot {
         SmartDashboard.putNumber("Pigeon Yaw", DriveBase.getInstance().getCurrentAngle());
         CommandScheduler.getInstance().run();
 
-        //The following 2 methods run Drivebase methods that tell shuffleboard what the motor current draw is and if motor current limiting is active.
-        DriveBase.getMotorCurrent();
-        DriveBase.isMotorLimitActive();
+        //The following 2 lines run Drivebase methods that tell shuffleboard what the motor current draw is and if motor current limiting is active.
+        DriveBase.getInstance().getMotorCurrent();
+        DriveBase.getInstance().isMotorLimitActive();
     }
 
     /**
@@ -198,10 +198,10 @@ public class Robot extends TimedRobot {
     @Override
     public void teleopPeriodic() {
         //If motor current limiting is active, trigger driver feedback (notify driver that drivetrain power is reduced).
-        if (DriveBase.isMotorLimitActive() == true) {
+        if (DriveBase.getInstance().isMotorLimitActive() == true) {
             //TODO: Put driver feedback code here. Consult with drive team's preferences.
             
-            //Example driver feedback (if you wish to use this, import joystick and RumbleType libraries)
+            //Example driver feedback (if you wish to use this, import joystick and RumbleType libraries):
             //joystick.setRumble(RumbleType.kLeftRumble, 0.5);
             //joystick.setRumble(RumbleType.kRightRumble, 0.5);
         } else {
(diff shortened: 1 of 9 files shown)


src/main/java/frc/robot/Robot.java
```diff
@@ -7,6 +7,10 @@
 
 package frc.robot;
 
+import java.io.PrintWriter;
```
georgetzavelas: Any reason to move these? Could be a setting in your IDE that is doing this?

src/main/java/frc/robot/Robot.java
```diff
@@ -40,6 +40,7 @@
     //flag to indicate the real match
     private Boolean bRealMatch;
 
+
```
georgetzavelas: Can you remove this extra line?

src/main/java/frc/robot/Robot.java
```diff
@@ -193,6 +198,12 @@ public void autonomousPeriodic() {
      */
     @Override
     public void teleopPeriodic() {
+        //If motor current limiting is active, trigger driver feedback (notify driver that drivetrain power is reduced).
+        if (DriveBase.isMotorLimitActive() == true) {
+            //TODO: Put driver feedback code here. Consult with drive team's preferences.
```
georgetzavelas, line 202: Can you put something here as a starting point?

src/main/java/frc/robot/Robot.java
```diff
@@ -193,6 +198,12 @@ public void autonomousPeriodic() {
...
+        } else {
+            //Disable current limit driver feedback here.
```
georgetzavelas: Should we be disabling here?

src/main/java/frc/robot/config/Config.java
```diff
@@ -141,7 +141,15 @@ private Config() {
             .registerToTable(Config.constantsTable);
     public static FluidConstant<Double> maxYawErrorOuterPortCommand = new FluidConstant<>("Outer Port Command Yaw Error", 3.0)
             .registerToTable(Config.constantsTable);
-    
+
+    // Current limiter values
+    public static final int kPeakCurrentAmps = 60; //Peak current threshold to trigger the current limit
```
georgetzavelas: To follow the convention within the code can you use capitals to define these and remove the `k` prefix?

src/main/java/frc/robot/subsystems/DriveBase.java
```diff
@@ -35,7 +36,16 @@
     private DriveMode driveMode;
 
     // The drivebase talons
-    private WPI_TalonSRX leftFrontTalon, leftRearTalon, rightFrontTalon, rightRearTalon, talon5plyboy;
+    private static WPI_TalonSRX leftFrontTalon, leftRearTalon, rightFrontTalon, rightRearTalon, talon5plyboy;
```
georgetzavelas: Since Drivebase is a singleton we should use a pattern that does not require static variables or static methods. e.g. always using `DriveBase.getInstance()`.

src/main/java/frc/robot/subsystems/DriveBase.java
```diff
@@ -68,6 +83,29 @@ private DriveBase() {
...
+        motorCurrent = DriveBase.leftFrontTalon.getSupplyCurrent();
+
+        SmartDashboard.putNumber("MotorCurrent (FrontLeft)", motorCurrent);
+        
+        return(motorCurrent);
+    }
+
+    public static boolean isMotorLimitActive() {
+        //Check if motor current limiting is active (is current draw over or at current limit)
+        if (motorCurrent > Config.kContinCurrentAmps - 1) {
```
georgetzavelas: Is -1 here a tolerance? Is there a reason not to use simply `motorCurrent > Config.kContinCurrentAmps`?

georgetzavelas (commented):
Great start Charlie! I think the extra files will disappear once you rebase. Some comments for you to consider.

src/main/java/frc/robot/Robot.java
Citysurvivor: OK.

src/main/java/frc/robot/Robot.java
Citysurvivor: Yes, driver feedback should be turned off once the current limiter stops doing its job.
I'll rewrite this comment to make this more clear.

Is that what you meant to say @georgetzavelas ?

src/main/java/frc/robot/config/Config.java
Citysurvivor: OK, fixed that.

src/main/java/frc/robot/subsystems/DriveBase.java
Citysurvivor: @georgetzavelas Yes, the `-1` is a tolerance, if I remember correctly what "tolerance" means. When the motor controller limits current draw, the current draw fluctuates, so I subtract 1 from the limit for a safety margin.

Should I place a comment in the code explaining what the `-1` is for?

georgetzavelas: If it is a tolerance then it would be best to define it as a constant so that the name of the constant makes it clear what its being used for 😄

src/main/java/frc/robot/Robot.java
Citysurvivor: No, I'm not sure what caused this. I'm not sure if the order makes a difference, but I'll put them back in the original order anyway.

nrgill28: This is default behavior in IntelliJ, it organizes imports alphabetically.

src/main/java/frc/robot/Robot.java
Citysurvivor: OK, will commit new code to provide an example.

src/main/java/frc/robot/subsystems/DriveBase.java
Citysurvivor: OK. Fixed this, although I might remove it in my next commit (if testing shows that I don't need it).

src/main/java/frc/robot/Robot.java
georgetzavelas: I mean that I would like to see some code here instead of a comment, even if it is a best guess.

Citysurvivor: OK, I'll fix that in my next commit.

src/main/java/frc/robot/subsystems/DriveBase.java
Citysurvivor: OK, fixed that. Committed new code, awaiting feedback.

Citysurvivor:
Closed. My branch was way behind master, so I rewrote my code on a new, up-to-date branch. See Motor Current Limiter 2.

PR #119, opened 2020-02-29
Lara shooter subsystem tuning

LaraLim:
I changed the indentation from 2 spaces to 4, so almost everything appears to have changed, but the actual major change is that the shooter will now ramp down instead of using closed loop velocity control to reach 0 RPM. The subsystem also prints the shooter motor temperature and current draw to shuffleboard.


src/main/java/frc/robot/commands/SpinUpShooter.java
```diff
@@ -1,51 +1,49 @@
...
+    // Called once the command ends or is interrupted.
+    @Override
+    public void end(boolean interrupted) {
+        shooterSubsystem.setTargetRPM(0);
+    }
+
+    // Returns true when the command should end.
+    @Override
+    public boolean isFinished() {
+        if (!shooterSubsystem.isActive()) {
```
Zinka010: Could change to return !shooterSubsystem.isActive()

Zinka010 (approved):
One small nitpick but overall a very good!

PR #121, opened 2020-02-29
Drive with Time command complete

IshanBaliyan:
Tested and completed driving with distance at any speed.


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -50,6 +51,10 @@
     private Command sensitiveDriverControlCommand;
     private Logger logger = Logger.getLogger("RobotContainer");
 
+    private final double RIGHT_SPEED = 0.2;
```
Zinka010: This should be in the constructor of the command, not here

src/main/java/frc/robot/commands/DriveWithDistance.java
```diff
@@ -0,0 +1,382 @@
...
+
+      }
+    }
+    return commandFinished;
+  }
+
+    // Called once the command ends or is interrupted.
+    @Override
+    public void end(boolean interrupted) {
+        SmartDashboard.putBoolean("Running", false);
```
Zinka010: Either remove this or make it more descriptive

Zinka010 (changes requested):
So it seems mostly ok with a few minor tweaks, though in the nicest way possible, why is it so complicated?

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -108,14 +113,20 @@ public Command getAutonomousCommand() {
             return null;
         }
 
+         else if (selectorOne == 1) {
```
georgetzavelas: Should this not be paired with selector two? Otherwise regardless of selector two this will be executed when selector one == 1.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -108,14 +113,20 @@ public Command getAutonomousCommand() {
...
+         /*
+          * When the selector is set to one, the robot will run for x seconds at y left motor speed and z right motor speed
+          */
 
-        if (selectorOne == 1 || selectorTwo == 1) {
-        /*
-         * When the selector is set to one, the robot will run for x seconds at y left motor speed and z right motor speed
-         */
+              // for the inputed variables: DriveWithTime(seconds (double), left motors speed (double), right motors speed (double))
+              return new DriveWithTime(1.0, 0.2, 0.2);
```
georgetzavelas: These are magic numbers. Can we use constants for these that get defined in Config?

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -108,14 +113,20 @@ public Command getAutonomousCommand() {
...
+         }
+
+        else if(selectorOne == 2){
+
+            //List distance, then the drive unit (in option of meters, cm, inches or feet), and then the right and left speed (if not specified, it is 0.5)
+            return new DriveWithDistance(0, 1, DEFAULT_UNIT, 0, 0.3);
```
georgetzavelas: Same here for magic numbers

src/main/java/frc/robot/commands/DriveWithDistance.java
```diff
@@ -0,0 +1,382 @@
...
+import frc.robot.subsystems.DriveBase.DistanceType;
+
+
+public class DriveWithDistance extends CommandBase {
+
+//Initializing private version of drivebase object
+private final DriveBase driveBase;
+
+// initialization of the variables 
+  public double currentRightDistance = 0;
```
georgetzavelas: I agree with Justin, when I see this many variables it tells me that this method is too complex. Maybe we can simplify this together?

src/main/java/frc/robot/commands/DriveWithDistance.java
```diff
@@ -0,0 +1,382 @@
...
+  public boolean commandFinished = false;
+  public boolean isTurning = false;
+  public boolean end = false;
+
+  /**
+   * Constructor to be used when the input specifications do not provide the right and left speed (assumed to be 0.5)
+   * 
+   * @param distance The distance to drive
+   * @param currentType The distance unit type to drive in
+   * @param rightSpeed The right motor speed
```
georgetzavelas: Extra params

src/main/java/frc/robot/commands/DriveWithDistance.java
```diff
@@ -0,0 +1,382 @@
...
+   * @param leftSpeed The left motor speed
+   */
+
+  public DriveWithDistance(double distance, DistanceType currentType) {
+
+    //If you have to move backwards
+    if(distance < 0){
```
georgetzavelas: 0 is a magic number here

src/main/java/frc/robot/commands/DriveWithDistance.java
```diff
@@ -0,0 +1,382 @@
...
+      //Default is meters
+      default:
+
+        this.currentConversion = DriveBase.DistanceUnit.METERS.encoderTicks;   
+        break;
+        
+    }
+
+    //default speed is 0.5 for the left and right talons
+    rightSpeed = 0.5;
```
georgetzavelas: Magic numbers

src/main/java/frc/robot/commands/DriveWithDistance.java
```diff
@@ -0,0 +1,382 @@
...
+      
+    }
+
+    //Setting current distance unit
+    this.currentType = currentType;
+
+    addRequirements(driveBase);
+    
+    //Setting the current conversion type according to the chosen distance unit
+    switch (currentType) {
```
georgetzavelas: This is the same as the previous constructor and the next one, it should be moved to a method and called by all the constructors.

src/main/java/frc/robot/commands/DriveWithDistance.java
```diff
@@ -0,0 +1,382 @@
...
+   * 
+   * @param distance The distance to drive
+   * @param currentType The distance unit type to drive in
+   * @param rightSpeed The right motor speed
+   * @param leftSpeed The left motor speed
+   */
+  public DriveWithDistance(double rightDistance, double leftDistance, DistanceType currentType, double rightSpeed, double leftSpeed) {
+    isTurning = true;
+
+    if(rightDistance < 0 && rightSpeed < 0){
```
georgetzavelas: Magic numbers

src/main/java/frc/robot/commands/DriveWithDistance.java
```diff
@@ -0,0 +1,382 @@
...
+      // Update to the current right/left distance that the robot has driven
+      currentRightDistance = DriveBase.getInstance().getRightDistance();
+      currentLeftDistance = DriveBase.getInstance().getLeftDistance();
+
+      // The robot is moving forward
+      if (!isTurning) {
+        // If the robot has reached or surpassed the desired distance, then stop the
+        // robot. Otherwise, keep moving (moving forward).
+        if (!backWards && currentRightDistance >= desiredRightDistance) {
+          driveBase.tankDrive(0, 0, false);
```
georgetzavelas: magic numbers

src/main/java/frc/robot/commands/DriveWithDistance.java
```diff
@@ -0,0 +1,382 @@
...
+          commandFinished = true;
+
+        }
+        // If the robot has reached or surpassed the desired distance, then stop the
+        // robot. Otherwise, keep moving (moving backward).
+        else if (backWards && currentRightDistance <= desiredRightDistance) {
+          driveBase.tankDrive(0, 0, false);
```
georgetzavelas: magic numbers

src/main/java/frc/robot/commands/DriveWithDistance.java
```diff
@@ -0,0 +1,382 @@
...
+  public double desiredRightDistance = 0;
+  public double encoderTicks;
+  public double rightEncoderTicks, leftEncoderTicks;
+  public double leftSpeed = 0;
+  public double rightSpeed = 0;
+  public DistanceType currentType;
+  public double currentConversion;
+  public boolean rightBackwards = false;
+  public boolean leftBackwards = false;
+  public boolean backWards = false;
```
georgetzavelas: This name is confusing, how is it different from right and left backwards? How can we incorporate that into the name?

src/main/java/frc/robot/commands/DriveWithDistance.java
```diff
@@ -0,0 +1,382 @@
...
+// initialization of the variables 
+  public double currentRightDistance = 0;
+  public double currentLeftDistance = 0;
+  public double desiredLeftDistance = 0;
+  public double desiredRightDistance = 0;
+  public double encoderTicks;
+  public double rightEncoderTicks, leftEncoderTicks;
+  public double leftSpeed = 0;
+  public double rightSpeed = 0;
+  public DistanceType currentType;
```
georgetzavelas: Since this is the only distance type, calling this current is kinda self evident. Why not distanceType?

src/main/java/frc/robot/config/Config.java
```diff
@@ -4,6 +4,7 @@
 import edu.wpi.first.networktables.NetworkTableInstance;
 import edu.wpi.first.wpilibj.DriverStation;
 import frc.robot.Robot;
+import frc.robot.subsystems.DriveBase;
```
georgetzavelas: Any reason this is needed in this class?

src/main/java/frc/robot/commands/DriveWithDistance.java
```diff
@@ -0,0 +1,382 @@
...
+    if(rightDistance < 0 && rightSpeed < 0){
+        rightBackwards = true;
+    }
+    else if(rightDistance > 0 && rightSpeed > 0){
+        rightBackwards = false;
+    }
+    //The robot should not move if it has inverse speeds and desired distances since that must be an error
+    else{
+      DriveBase.getInstance().tankDrive(0, 0, false);
+      commandFinished = true;
```
georgetzavelas: I am not seeing anywhere that would read `commandFinished` so it should not be a class variable and instead a local variable to `isFinished()`

src/main/java/frc/robot/commands/DriveWithDistance.java
```diff
@@ -0,0 +1,382 @@
...
+        }
+
+      }
+      // If the robot is turning, then move the robot for the left side too
+      else {
+
+        // If the robot has reached or surpassed the desired distance, then stop the
+        // robot. Otherwise, keep moving (moving forward).
+        if (!rightBackwards && currentRightDistance >= desiredRightDistance) {
+          rightSpeed = 0;
```
georgetzavelas: This condition and the next one both have the same result, can you combine them?

src/main/java/frc/robot/commands/DriveWithDistance.java
```diff
@@ -0,0 +1,382 @@
...
+        else if (rightBackwards && currentRightDistance <= desiredRightDistance) {
+           rightSpeed = 0;
+          
+        } else {
+          commandFinished = false;
+        }
+
+        // If the robot has reached or surpassed the desired distance, then stop the
+        // robot. Otherwise, keep moving (moving forward).
+        if (!leftBackwards && currentLeftDistance >= desiredLeftDistance) {
```
georgetzavelas: This condition and the next one both have the same result, can you combine them?

src/main/java/frc/robot/commands/DriveWithDistance.java
```diff
@@ -0,0 +1,382 @@
...
+//Initializing private version of drivebase object
+private final DriveBase driveBase;
+
+// initialization of the variables 
+  public double currentRightDistance = 0;
+  public double currentLeftDistance = 0;
+  public double desiredLeftDistance = 0;
+  public double desiredRightDistance = 0;
+  public double encoderTicks;
+  public double rightEncoderTicks, leftEncoderTicks;
```
georgetzavelas: These feel more like desiredRightEncoderTicks and desiredLeftEncoderTicks

src/main/java/frc/robot/commands/DriveWithDistance.java
```diff
@@ -0,0 +1,382 @@
...
+
+  
+  
+  /**
+   * Determines when the command is finished using the current distnce and desired distance
+   * 
+   */
+
+  @Override
+  public boolean isFinished() {
```
georgetzavelas, line 193: This function should only determine if the command has finished, it should not alter variables, that should be done in execute().

src/main/java/frc/robot/commands/DriveWithDistance.java
```diff
@@ -0,0 +1,382 @@
...
+  public void initialize() {
+
+    //Add the desired amount of encoder ticks to the current distance to get the amount of encoder ticks that the robot needs to drive
+    desiredRightDistance = driveBase.getRightDistance() + rightEncoderTicks;
+    desiredLeftDistance = driveBase.getLeftDistance() + leftEncoderTicks;
+
+    addRequirements(DriveBase.getInstance());
+
+    //Dashboard
+    SmartDashboard.putBoolean("Running", true);
```
georgetzavelas: What is running?

src/main/java/frc/robot/commands/DriveWithDistance.java
georgetzavelas: Can we not simply have this constructor call the next one instead of duplicating the logic? e.g. DriveWithDistance(distance, currentType, DEFAULT_LEFT_SPEED, DEFAULT_RIGHT_SPEED)

georgetzavelas (commented):
You did a lot of great work here Ishan! I think there is some great opportunities here to refactor this to simply it without loosing your logic. I left you some suggestions.

georgetzavelas:
You should also rebase before making your changes.

src/main/java/frc/robot/RobotContainer.java
IshanBaliyan: I put it in RobotContainer because they are values you are supposed to call on line 128 since I don't want to use "Magic Numbers". I'll add them to config and then reference them on line 128.

src/main/java/frc/robot/RobotContainer.java
IshanBaliyan: Yes, you are correct. However, on Mergio the robot's second selector is broken so we decided to tailor the code to one selector. Furthermore, I just learned that the 2020 robot will have one selector so I will just remove the code for the second selector now.

src/main/java/frc/robot/RobotContainer.java
IshanBaliyan: Done. Added them to config and referenced them here.

src/main/java/frc/robot/RobotContainer.java
IshanBaliyan: Done. Added them to config and referenced them here.

src/main/java/frc/robot/commands/DriveWithDistance.java
IshanBaliyan: Renamed variables.

src/main/java/frc/robot/commands/DriveWithDistance.java
IshanBaliyan: Renamed to distanceType.

src/main/java/frc/robot/commands/DriveWithDistance.java
IshanBaliyan: Right backwards is for the right motor moving backwards. Left backwards is for the left motor moving backwards. The motors could be moving in opposite directions when turning. The regular variable for "backwards" is used for when both motors are moving backwards simultaneously.  I renamed the variables to better suit their purpose.

src/main/java/frc/robot/commands/DriveWithDistance.java
IshanBaliyan: Fixed

src/main/java/frc/robot/commands/DriveWithDistance.java
IshanBaliyan: Yeah, that is an amazing idea! Never thought of that way. Will change that now.

src/main/java/frc/robot/commands/DriveWithDistance.java
IshanBaliyan: Declared variable and replaced magic number.

src/main/java/frc/robot/commands/DriveWithDistance.java
IshanBaliyan: Fixed

src/main/java/frc/robot/commands/DriveWithDistance.java
IshanBaliyan: I'm not entirely sure, but It is used in the constructors so I think it needs to be a class variable so it can be accessed in isFinished() and the constructors.

src/main/java/frc/robot/commands/DriveWithDistance.java
IshanBaliyan: Removed.

src/main/java/frc/robot/commands/DriveWithDistance.java
IshanBaliyan: Fixed

src/main/java/frc/robot/commands/DriveWithDistance.java
IshanBaliyan: fixed

src/main/java/frc/robot/commands/DriveWithDistance.java
IshanBaliyan: Fixed

src/main/java/frc/robot/commands/DriveWithDistance.java
IshanBaliyan: Fixed

src/main/java/frc/robot/config/Config.java
IshanBaliyan: Yes, it is used with a private final version of drivebase on line 19

src/main/java/frc/robot/commands/DriveWithDistance.java
IshanBaliyan: Removed

src/main/java/frc/robot/commands/DriveWithDistance.java
IshanBaliyan: moved to execute

src/main/java/frc/robot/commands/DriveWithDistance.java
```diff
@@ -0,0 +1,196 @@
...
+    //Run motors according to output of the speeds
+    driveBase.tankDrive(leftSpeed, rightSpeed, false);
+
+    //Dashboard
+    SmartDashboard.putNumber("Left Speed", leftSpeed);
+    SmartDashboard.putNumber("Right Speed", rightSpeed);
+    SmartDashboard.putNumber("Left Error", desiredLeftDistance - currentLeftDistance);
+    SmartDashboard.putNumber("Right Error", desiredRightDistance - currentRightDistance);
+
+    //Code for figuring out when the robot is finished driving and handeling user input validation
```
nicojopr: minor typo on handling

src/main/java/frc/robot/commands/DriveWithDistance.java
```diff
@@ -0,0 +1,196 @@
...
+
+    // Update to the current right/left distance that the robot has driven
+    currentRightDistance = DriveBaseHolder.getInstance().getRightDistance();
+    currentLeftDistance = DriveBaseHolder.getInstance().getLeftDistance();
+
+    // If the robot has reached or surpassed the desired distance, then stop the
+    // robot. Otherwise, keep moving (moving forward OR moving backwards).
+    if (Math.abs(currentRightDistance) >= desiredRightDistance) {
```
nicojopr, line 167: Question: as soon as either left or right are 'far enough' wouldn't we want to stop both anyway to avoid turning?

nicojopr (commented):
You should look at latest on master as there is a DriveBase2020 so I'm not sure which is being used v.s. Pre2020 and plain DriveBase. Master is no longer compatible with the older bots.

Also, there are still merge conflicts to resolve

src/main/java/frc/robot/commands/DriveWithDistance.java
IshanBaliyan: No, we would want it to turn in that scenario. Depending on the constructor the user uses, the robot may have a different "desired distance" for the left and right motor.

IshanBaliyan:
> You should look at latest on master as there is a DriveBase2020 so I'm not sure which is being used v.s. Pre2020 and plain DriveBase. Master is no longer compatible with the older bots.
> 
> Also, there are still merge conflicts to resolve

Yes, I changed some code to work with the new drivebase but it needs to be tested with the new robot before I push it.

georgetzavelas:
Any reason this got closed?

IshanBaliyan:
Sorry, will reopen. Must have closed by accident.

nicojopr (approved):

Zinka010:
So I'll approve this for now but I need you to understand that there are much better ways of doing drive straight commands since CTRE gives us so many great tools. I'd really like it if someone could get something like this for North Bay https://github.com/CrossTheRoadElec/Phoenix-Examples-Languages/blob/master/Java%20General/DriveStraight_AuxPigeon/src/main/java/frc/robot/Robot.java.

What you're doing, while functional I'm sure is just running for a set amount of encoder ticks then stopping. As a next step, we can use the PigeonIMU to ensure we are keeping straight throughout this movement. The robot characterization tool is great for getting PID values for this, though note we'd NEED someone to come to a practice field or Entrust to do it.

Zinka010 (approved):

IshanBaliyan:
> So I'll approve this for now but I need you to understand that there are much better ways of doing drive straight commands since CTRE gives us so many great tools. I'd really like it if someone could get something like this for North Bay https://github.com/CrossTheRoadElec/Phoenix-Examples-Languages/blob/master/Java%20General/DriveStraight_AuxPigeon/src/main/java/frc/robot/Robot.java.
> 
> What you're doing, while functional I'm sure is just running for a set amount of encoder ticks then stopping. As a next step, we can use the PigeonIMU to ensure we are keeping straight throughout this movement. The robot characterization tool is great for getting PID values for this, though note we'd NEED someone to come to a practice field or Entrust to do it.

I will work on that for North Bay

PR #124, opened 2020-03-01
Charlie current limiter 2 (Issue #13)

### LINKED ISSUE #13: Prototyping: Current Limiting on Drivetrain

Citysurvivor:
Added driver feedback (joystick rumble) for motor current limiter code. Fixed comments for current limiter code.

### DIFF
--- a/src/main/java/frc/robot/RobotContainer.java
+++ b/src/main/java/frc/robot/RobotContainer.java
@@ -39,8 +39,7 @@ public class RobotContainer {
    * The container for the robot. Contains subsystems, OI devices, and commands.
    */
     
-    // RobotContainer is a singleton class as there is only one robot
-    //TODO: CHECK THIS
+    // RobotContainer is a singleton class
     private static RobotContainer currentInstance;
 
   // The robot's subsystems and commands are defined here...    
--- a/src/main/java/frc/robot/RobotContainer.java
+++ b/src/main/java/frc/robot/RobotContainer.java
@@ -10,6 +10,7 @@ package frc.robot;
 import edu.wpi.first.wpilibj.GenericHID;
 import edu.wpi.first.wpilibj.Joystick;
 import edu.wpi.first.wpilibj.XboxController;
+import edu.wpi.first.wpilibj.GenericHID.RumbleType;
 import edu.wpi.first.wpilibj2.command.Command;
 import edu.wpi.first.wpilibj2.command.button.JoystickButton;
 import frc.robot.commands.*;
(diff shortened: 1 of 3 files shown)


Citysurvivor:
Temporarily closed - I am making changes tonight.

Zinka010 (approved):
👍

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -134,5 +138,26 @@ else if (selectorOne == 1 || selectorTwo == 1) {
...
+    /**
+     * Initialize the current RobotContainer instance
+     */
+    public static void init() {
+        if (currentInstance == null) {
+            currentInstance = new RobotContainer();
+        }
+    }
+
+    public static RobotContainer getInstance() {
```
nicojopr, line 157: If you decide to change RobotContainer implementation to follow the singleton pattern then you should make the basic constructor (public RobotContainer() ) private, ensure you change the current creation of RobotContainer  to use the  getInstance (or we will have multiple RobotContainers).

To keep your changes simple, perhaps leave RobotContainer initialization as is and just add the joystick rumble parts

src/main/java/frc/robot/subsystems/DriveBase2020.java
```diff
@@ -58,7 +58,7 @@ public double getMotorCurrent() {
     public boolean isMotorLimitActive() {
         //Check if motor current limiting is active (is current draw over or at current limit)
         if (motorCurrent >= Config.CONTIN_CURRENT_AMPS) {
-            motorLimitActive = true;
+            motorLimitActive = true; //For driver feedback purposes only.
```
nicojopr, line 61: Does this mean the drivetrain current limiting is on by default?

src/main/java/frc/robot/RobotContainer.java
Citysurvivor: True, that's a good point. No need to reinvent what has already been done. I will undo my changes and use the existing RobotContainer initialization.

src/main/java/frc/robot/subsystems/DriveBase2020.java
Citysurvivor: No, motorLimitActive is meant for driver feedback, not current limiting.

motorLimitActive is false by default. motorLimitActive will be set to true if motorCurrent rises above CONTIN_CURRENT_AMPS (the continuous current limit).

I'll rewrite my comments to improve clarity.

PR #130, opened 2020-03-07
Tuned shooter to decent accuracy

LaraLim:
- 1300 RPM 
- 30% feeder output
- P = 0.0025
- I = 0.0
- D = 0.004
- F = 0.0002

### DIFF
--- a/src/main/java/frc/robot/commands/VisionAssistedShooter.java
+++ b/src/main/java/frc/robot/commands/VisionAssistedShooter.java
@@ -1,3 +1,4 @@
+package frc.robot.commands;
 
 import frc.robot.config.Config;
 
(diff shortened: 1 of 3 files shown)


nicojopr (approved):

src/main/java/frc/robot/commands/VisionAssistedShooter.java
```diff
@@ -1,3 +1,4 @@
+package frc.robot.commands;
```
georgetzavelas, line 1: Do you know why this line is needed when nothing else changed in the file?

georgetzavelas (commented):
My favourite question, can we make these into constants? :)

src/main/java/frc/robot/commands/VisionAssistedShooter.java
LaraLim: When I pulled from master it was somehow missing, so it caused an error.

Zinka010 (approved):

georgetzavelas:
Maybe we can circle back when people have more time and make these into constants based on my comment? :)

PR #134, opened 2020-03-09
Shooter tuning

Zinka010:
Further tuning of the competition robot robot

### DIFF
--- a/src/main/java/frc/robot/Robot.java
+++ b/src/main/java/frc/robot/Robot.java
@@ -19,6 +19,7 @@ import frc.robot.nettables.VisionCtrlNetTable;
 import frc.robot.subsystems.DriveBase2020;
 import frc.robot.subsystems.DriveBaseHolder;
 import frc.robot.sensors.AnalogSelector;
+import frc.robot.subsystems.ShooterSubsystem;
 
 import java.io.PrintWriter;
 import java.io.StringWriter;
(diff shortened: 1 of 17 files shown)


georgetzavelas (approved):
That's a lot of changes to commit in one go! Moving forward I suggest to try and raise PRs more often for new changes so that we can isolation changes. Otherwise if something ends up not working, you may have to end up reverting all of these changes to find the culprit.

I see some magic numbers, maybe make a note to come back and fix them after Carleton (or tag the owners to fix now)? e.g. `MoveArmManuallyCommand(-0.35)` or `if (currentTarget != -99)`, I have no idea what these means and they are likely code from others that you are simply modifying.

Zinka010:
Yeah, I realize my PR's are way too big, but its mostly just because I just end up making them at the end of a day of tuning many different things. I'm not sure if there's an efficient way to make multiple changes that rely on each other and also be able to quickly switch what I'm working on based on a shift in drive team and mentor requests. I'd appreciate any advice on it since I agree that this makes issues much harder to pinpoint.

georgetzavelas:
Totally fair and if this is what works best for you to handle the chaos then I support you fully!

One way I handle changes that depend on each other is to create cascading branches, e.g. master -> isolated_change_1 -> isolated_change_2 -> isolated_change_3, where each branch has individual changes. Then you merge isolated_change_1 to master first, after it is merged, isolated_change_2 base branch can change to master, and so on. I can show you what I mean on Wed if you want to learn more.

Zinka010:
Maybe not Wednesday since I'm still swamped with stuff to do but I'd love to after Carleton

PR #138, opened 2020-03-12
Incrmenenting stop at sensor

Zinka010:
- Fixed broken driving code caused by current limiting
- Added manual arm positioning


nicojopr (approved):
Awesome integration work! Glad to have the latest merged. At some point, I'd like to see some of the magic numbers added as constants in Config.java. In addition, comments on why the current limiting code has been commented out would be helpful.

